PR #3, opened 2018-08-30
Reintegrate my changes from the original version of statick, add some templates, fix bandit plugin

creffett:
All changes are released as public domain.


creffett:
(will mark as not WIP once I've finished testing these changes)

creffett:
Ready to merge!

.travis.yml
```diff
@@ -12,7 +12,7 @@ notifications:
 before_install:
   - sudo apt-get update
     #- cat install.txt | xargs sudo apt-get install -y
-  - sudo apt-get install -y clang-3.9 clang-format-3.9 clang-tidy-3.9 flawfinder libxml2 pylint python-git python-yapsy
+  - sudo apt-get install -y clang-3.9 clang-format-3.9 clang-tidy-3.9 flawfinder libxml2 pylint python-git python-yapsy findbugs flawfinder
```
denewiler: `flawfinder` is listed twice now.

statick_tool/plugins/tool/make_tool_plugin.py
```diff
@@ -101,10 +101,13 @@ def parse_output(self, package, output):
...
         issues = []
         for match in matches:
+            cert_reference = None
             warning_list = warning_parse.match(match[4])
+            if warning_list is not None and warning_list.groups(1)[0] in warnings_mapping.keys():
+                cert_reference = warnings_mapping[warning_list.groups(1)[0]]
 
             if warning_list is None:
-                # Something's gone wrong if we don't match the [warning] format
+                # Something's gone wrong if we don't match the [waring] format
```
denewiler: This just added a typo.

creffett: So it did. Not sure where that came from, will fix.

.travis.yml
creffett: Good catch!

statick_tool/plugins/tool/clang_format_tool_plugin.py
```diff
@@ -59,7 +59,7 @@ def scan(self, package, level):
             output = subprocess.check_output([clang_format_bin,
                                               "--dump-config"],
                                              stderr=subprocess.STDOUT)
-            format_file_name = self.plugin_context.resources.get_file(".clang-format")
+            format_file_name = self.plugin_context.resources.get_file("_clang-format")
```
denewiler: Why do you want to switch from `.clang-format` to `_clang-format`? They appear to still have the same contents. I like `.clang-format` because you can copy it straight into your home directory and apply it to files and know that they will pass the statick check.

creffett: That's a holdover from the changes I made originally, I seem to recall having some issue with .clang_format vs _clang_format but I don't remember the specifics. I'll remove that change and see if it still works.

denewiler: Okay. If it doesn't work just put that change back and I'll open a new issue to investigate.

denewiler:
The default levels work for me. I'll add some fixes after this for a few tools that do not work.

PR #7, opened 2018-08-31
Add Sphinx documentation auto-generation in Travis

creffett:
Will need some configuration on Travis to let it auto-push the gh-pages branch, see https://github.com/Syntaf/travis-sphinx#obtaining-a-personal-access-token.


.travis.yml
```diff
@@ -14,7 +14,12 @@ before_install:
     #- cat install.txt | xargs sudo apt-get install -y
   - sudo apt-get install -y clang-3.9 clang-format-3.9 clang-tidy-3.9 flawfinder libxml2 pylint python-git python-yapsy findbugs
   - sudo -H pip install -r requirements.txt
+  - sudo pip install sphinx travis-sphinx
```
denewiler: These should be added to `requirements.txt`.

denewiler:
I added a personal access token for gh-pages and linked it to the Travis job. We should be good on that side.

.travis.yml
denewiler: Actually, I can see these being separate since they are only used by Travis, not by typical developers.

tdenewiler:
@creffett, looking good.

https://sscpac.github.io/statick/

PR #16, opened 2018-09-12
Handle missing binaries

tdenewiler:
This can be tested on a local developer machine by removing pep257 ( I used `sudo apt-get remove pep257`), going to the statick directory, and running `./statick . statick_output --profile self_check.yaml`. Without these changes statick should crash and not complete running. With these changes statick should print out a message saying that the tool failed to run, give a recommendation to install the tool, and then continue on to the next tool -- you should see a full report at the end.


denewiler:
There were several cases where Statick would crash if a tool was missing. This PR will now attempt to address all missing tools so that Statick reports failure for missing tools but continues gracefully for tools that are available.

denewiler:
Something weird is happening with Travis CI builds. The current master branch is now failing, and it is failing on the `travis-sphinx build` command. See https://travis-ci.org/tdenewiler/statick/builds/427989168. There are no statick issues reported. I'm going to approve this PR since it has the same cause for failure as the master branch.

Issue #45, opened 2019-01-17
Findbugs quietly failing due to argument handling problems


creffett:
Running Statick in an Ubuntu 16.04 environment, I got the following backtrace when running with --show-tool-output:
```
Running findbugs tool plugin...
[warning] /usr/bin/findbugs: Unable to locate ant in /usr/share/java
The following errors occurred during analysis:
  File not found: filesystem:/home/jenkins/workspace/ROSM_rtk-statick_master-XIVWAOJYVSYPP4AQ5FZIKKA755OLWXMYAKWA4FWMINIDWHGUV7YA/statick_output/dsat_jaus2ros_sensor-sei_cert/longBugCodes
  File not found: filesystem:/home/jenkins/workspace/ROSM_rtk-statick_master-XIVWAOJYVSYPP4AQ5FZIKKA755OLWXMYAKWA4FWMINIDWHGUV7YA/statick_output/dsat_jaus2ros_sensor-sei_cert/-include\
  Cannot open codebase filesystem:/usr/local/lib/python2.7/dist-packages/statick_tool/rsc/findBugs-security.xml
    java.io.IOException: Wrong magic bytes of 3c46696e for zip file /usr/local/lib/python2.7/dist-packages/statick_tool/rsc/findBugs-security.xml of 249 bytes
      At edu.umd.cs.findbugs.classfile.impl.ZipFileCodeBase.<init>(ZipFileCodeBase.java:87)
      At edu.umd.cs.findbugs.classfile.impl.ZipCodeBaseFactory.makeZipCodeBase(ZipCodeBaseFactory.java:39)
      At edu.umd.cs.findbugs.classfile.impl.ClassFactory.createFilesystemCodeBase(ClassFactory.java:121)
      At edu.umd.cs.findbugs.classfile.impl.FilesystemCodeBaseLocator.openCodeBase(FilesystemCodeBaseLocator.java:77)
      At edu.umd.cs.findbugs.classfile.impl.ClassPathBuilder.processWorkList(ClassPathBuilder.java:607)
      At edu.umd.cs.findbugs.classfile.impl.ClassPathBuilder.build(ClassPathBuilder.java:226)
      At edu.umd.cs.findbugs.FindBugs2.buildClassPath(FindBugs2.java:677)
      At edu.umd.cs.findbugs.FindBugs2.execute(FindBugs2.java:218)
      At edu.umd.cs.findbugs.FindBugs.runMain(FindBugs.java:402)
[...]

PR #47, opened 2019-01-17
Findbugs fixes for issue #45

creffett:
Fixes findbugs issues:
* Add missing mandatory -textui arg at the start of the args list
* Add missing prefix - to longBugCodes argument
* Split -include and -exclude file arguments into two argument entries (was being parsed as one big argument with a space in the middle)
* Skip the plugin if no java .class files were found


tdenewiler:
I am trying to find a Java repository that I can clone and test that findbugs runs without the errors you showed in #45. All the ones I found end up being discovered as Java source files, but no Java class files. Then it seems that findbugs is not running and I don't see any output. Do you have any recommendations?

creffett:
Yeah, I realized that - the issue is basically that the java files need to be compiled either before or during the Statick run, since findbugs only works on compiled Java. For now, what I did to test it was to make a "test" directory, drop the following file in there, manually compile (with java 8 - findbugs breaks with >= java 9), and then run Statick. 

Test file:
```
public class Test {
public static int bad = 1;
public static void main(String[] args) {
}
}
```

If you run Statick with --show-tool-output, you'll see that there is an error found in this file. It might not actually show in the output, but that's a different bug I'm about to file.

denewiler:
I saw the error in the test file you showed.

PR #52, opened 2019-01-24
Adding a check in the exceptions module to ensure the issue to be ign…

tdenewiler:
…ored has an absolute path to the filename containing the issue. Documentation in the source code and users guide was updated to make it more clear that file exceptions require an absolute path to the issue filename.

@creffett. I think this is the minimum of what you wanted to see. Let me know if you want more changes for this initial pass at addressing exceptions for findbugs.


creffett:
Looks good here, thanks for taking care of that!

PR #54, opened 2019-01-25
Use tox for code checking

creffett:
Most of these changes are things like import order sorting or pep8 cleanups so that the tox tests actually pass. I went with the following pytest plugins:
* pytest-cov
* pytest-flake8
* pytest-isort
* pytest-pep257

If there are any you don't like in there, I'm fine with changing that list, it was just based on the plugins I've used on other projects. Reason we're using tox + pytest is to lay the ground work for actual unit testing and multi-python tests in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,8 @@ dist/
 *.egg-info/
 
 docs/_build
+
+# From tests
+.coverage
+.pytest_cache
+.tox
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ docs/_build
 .coverage
 .pytest_cache
 .tox
+output-*/
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,6 @@ matrix:
     - python: 2.7
       env:
         - TOX_ENV=py27
-    - python: 3.5
-      env:
-        - TOX_ENV=py35
 
 jobs:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: required
 language: python
 python:
   - '2.7'
+
 cache:
   - apt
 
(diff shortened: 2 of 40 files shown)


tdenewiler:
I just did a quick look at the diff and nothing jumped out at me. All those changes to files look good. I'll review it more in depth soon. Looking at tox will take me a bit longer since I have not used it before. I'm excited to see a buildup to actual testing. Thanks!

creffett:
The general idea is that it spins up a virtualenv for each of several python versions (and I think it can also matrix with other versions, so you could do, say, several python version x several django versions for a webapp), and then runs tests and such for each of them. This way, we can test that Statick works with multiple Python versions in the future (and I should have a merge request for Python 3 compatibility pretty soon after this is merged)

tdenewiler:
I am having an issue with the cmake discovery plugin that affects the catkin-format tool plugin. I could likely make a more minimal example, but I am running against my [node_example](https://github.com/tdenewiler/node_example) package. I have that at `~/src/node_example/src/node_example`. For each of the master and tox branches I run the following:

    ./statick ~/src/node_example/src/node_example/ statick_output/ --user-paths ~/src/node_example/src/node_example/statick_config/ --force-tool-list=clang-format

The master branch is fine, but I get the following error on the tox branch:

```
Running clang-format tool plugin...
  File "./statick", line 34, in <module>
    main()
  File "./statick", line 24, in main
  File "/home/thomas/tmp/statick/statick_tool/statick.py", line 212, in run
    tool_issues = plugin.scan(package, level)
  File "/home/thomas/tmp/statick/statick_tool/plugins/tool/clang_format_tool_plugin.py", line 46, in scan
    flags = ["-header-filter=" + package["src_dir"] + "/.*", "-p",
KeyError: 'src_dir'
```

I grabbed the `statick_tool/plugins/discovery/cmake_discovery_plugin.py` file from master branch and copied that over the modified version in the tox branch. Running with that single reverted file in the tox branch I did not get any errors -- the output matched the output from running the master branch.

I have not looked into what the syntax changes in that file are doing yet. I'll try to investigate more later. If you have any ideas please let me know.

Everything else in this PR looks good to me so far. I'll run a couple more tests on my side to see if anything else pops up.

tdenewiler:
Note that I got the same clang-format errors when I ran without `--force-tool-list=clang-format`. It is just much faster to run with that flag.

creffett:
Got it, I'll have a look.

PR #62, opened 2019-01-29
Add initial unit tests

### LINKED ISSUE #10: Add unit tests
I'd say the highest priorities are adding tests for the various tools (so that we get a warning if their output format changes) and for the various file- and configuration-overlaying parts (making sure user directories correctly override default settings, testing profile inheritance, etc.). If that sounds good to you, I'll take a crack at it next month once I have the hours.

creffett:
Still a work in progress, not ready to merge, but this branch adds some basic unit tests. Fixes #10.


statick_tool/args.py
```diff
@@ -28,10 +28,10 @@ def __init__(self, name):
         self.parser = argparse.ArgumentParser(description=name)
         self.parser.add_argument("--user-paths", **user_path_args)
 
-    def get_user_paths(self):
+    def get_user_paths(self, args=None):
         """Get a list of user paths containing config or plugins."""
         user_paths = []
-        args = self.pre_parser.parse_known_args()[0]
+        args = self.pre_parser.parse_known_args(args)[0]
```
creffett: This is safe because the default signature of parse_known_args is (args=None, namespace=None), so if args is None, then the behavior is completely unchanged.

creffett:
The moment we've all been waiting for: unit testing!

Notes about this merge:
* All of the tox runs will have a warning about deprecated calls - those are in other libraries (e.g. isort) and I can't for the life of me figure out how to get rid of them
* Functionality changes to help testing:
  * Added catches for things like nonexistent 
  * Moved the shlex user flag parsing functionality to a common function in tool_plugin. Fixed an issue where if get_tool_config returned `None`, shlex would interpret that as "wait forever for command-line input," which made it hard to test certain cases and is probably not what we wanted anywya.
  * Added optionally passing an args object to get_args and get_user_paths (again, for testing, since we can't manipulate the command line of a test)
  * Have the cmake plugin set `package['cmake'] = False` if the package is not cmake, so that other plugins can rely on the value of package['cmake'] instead of checking `if 'cmake' in package.keys()`
  * Removed executable checking from the python discovery plugin - there's no guarantee that a python file has the executable bit set so there's no value in the check
  * Add universal_newlines to all of the subprocess calls - without it, Python 3 gets hung up on the plugin output type (something like bytes vs str)

tdenewiler:
This is fantastic.

I don't see any issues when I look through the diff. There are two things that I'm not sure about yet.

1. When I run `./statick . statick_output/ --profile self_check.yaml` locally I get a report with 383 unique issues. That same command is run in Travis and does not give the same output. Almost all of the reported issues are in the `tests` directory and seem like true positives. I do see one missing docstring issue for `statick_tool/tool_plugin.py`. Was there an exception for the `tests` directory added somewhere?

2. It would be nice to add some instructions to the README explaining how to run the unit tests locally. I would rather fix things locally before committing/pushing and having Travis fail my builds.

creffett:
That's very interesting...we definitely _should_ ignore the tests directory (or at least anything not matching test_*.py) since that's the whole point of having test files. Will investigate. Also will add to the README.

creffett:
I'm seeing a bunch of pylint true positives even in the statick_tool directory (and ones which have been there for a while). I suspect something's up with the Travis self_check run.

creffett:
Oh, this is wonderful. The path travis is running Statick in...is /home/travis/build/sscpac/statick...and we ignore */build/* in exceptions.yaml. I'll file a separate issue about that.

creffett:
Regarding the Travis failures: looks like some kind of API breakage involving flake8's newest release, it looks like they've merged a fix but haven't published a new version with that fix yet.

tdenewiler:
I'm good with merging even with the Travis flake8 failures, as long as that is going to be temporary. Looks like they might have already done the release of 3.7.2: https://pypi.org/project/flake8/. I think the API breakage you are referring to is the issue at: https://gitlab.com/pycqa/flake8/issues/497.

denewiler:
Nope, not fixed yet. But I can look at the Travis output to see if it is only failing due to flake8 false positives. If the failures continue for a long time we can look into other approaches.

tdenewiler:
FYI, I tested a branch on my fork where flake8 is removed. That passed in Travis, so it can be our fallback.

https://travis-ci.org/tdenewiler/statick/builds/486756395

Issue #51, opened 2019-01-22
Add maven discovery plugin


creffett:
Should be pretty straightforward, could either find the topmost pom.xml(s) or find all of them.

PR #72, opened 2019-02-05
Add Maven discovery plugin

creffett:


creffett:
Fixes #51

denewiler:
Travis is showing a successful build even though the PR is showing that the Travis check is still pending. I'm calling it good.

PR #83, opened 2019-02-20
Add support for a suffix to issue file names, fixes #81

### LINKED ISSUE #81: Add support for changing the issue file that is looked up
The idea is that there could be multiple mapping files (the use case here is that there are the "official" mapping files, based on the CMU-SEI published list, but we would like to add "unofficial" mappings which are not blessed by CMU-SEI). I'm thinking that the most straightforward approach is to support adding a suffix (e.g. findbugs-experimental.txt) and try to look that up if it exists. Shouldn't be too complicated.

creffett:


denewiler:
tox is passing locally for me, but there are errors on Travis. These errors look similar to some of the issues I was having with `statick_ws`, where the plugins were not getting arguments passed to them the way I would expect.

creffett:
Ahh, the issue is that the perl PR hadn't been accepted when I made the MR but was there after you made your edit, so it caught the fact that one tweak I had to make in the tests hadn't been made in the perlcritic test. Easy fix.

PR #84, opened 2019-02-21
Add CCCC tool plugin

tdenewiler:


tdenewiler:
d7db0a4 added ability to set CCCC binary path. The user args are listed at:

http://sarnold.github.io/cccc/CCCC_User_Guide.html#command

I am using a specific input file format for the configuration and a specific output file format to parse the results. I don't see much value in supporting user arguments for this tool.

Reading the user guide again I realized that an enhancement for the future would be to support Java files as well as C/C++ files.

Thanks for reviewing!

PR #85, opened 2019-02-21
Travis build exception

tdenewiler:
Adding hack to avoid ignoring all issues when building on Travis CI due to build directory used by Travis CI.


creffett:
Looks good to me.

Issue #65, opened 2019-01-30
Statick self test on Travis returns no results


creffett:
Appears to be because exceptions.yaml adds an exception for `*/build/*` and the build environment is located in /home/travis/build/sscpac/statick, so every issue trips the `*/build/*` exclude and is ignored.

tdenewiler:
I have a potential fix for this, but there are a couple of issues that cropped up.

1. If we want to be able to run the self test on Travis and have it fail the Travis build we will need to have the exit code be non-zero if any issues are discovered. Currently, if issues are found that does not necessarily set a non-zero exit code. I am okay with making that change, but wanted to see if you are too.

2. On Travis, due to installing via pip, we are getting `pylint>=1.8.2  (from pylint-django<2.0->-r requirements.txt (line 7))`. This is a different version of pylint than what is available from xenial apt repositories (1.5.2-1ubuntu1). When running the self check against this newer version of pylint there are additional warnings generated.

If we set a non-zero exit code when issues are identified, and we are okay using the newer version of pylint, then the fixes I have for getting the self test to work on Travis means we will have to fix the new issues in Statick. None of the new issues look difficult to fix, and I am happy to do it. I just want to make sure that you don't see any problems with that approach before I do the work.

tdenewiler:
Addressing this with #85 and #87.

tdenewiler:
Fixed by #87.

PR #87, opened 2019-02-22
Travis self check passes

tdenewiler:
* Changing pylint to install from apt (1.5.2-1) to pip (1.8.2).

  * Fixing pylint issues identified by using newer version from pip.

  * Setting exit code appropriately so that warnings from self check
    configuration will fail a Travis CI build.

  * Allowing some common single letter variable names to pass pylint in
    self check configuration.


examples/navigation/navigation_config/plugins/my_custom_tool_plugin.py
```diff
@@ -6,8 +6,8 @@
 import shlex
 import subprocess
 
-from statick_tool.issue import Issue
-from statick_tool.tool_plugin import ToolPlugin
+from statick_tool.issue import Issue  # pylint: disable=import-error
```
tdenewiler: It would be nice to fix these and not disable the warning.

statick
```diff
@@ -24,6 +24,9 @@ def main():
 
     path = parsed_args.path
     issues, success = statick.run(path, parsed_args)
+    for tool in issues:
```
tdenewiler: This is the bit that causes Statick to return an exit code if any of the tool plugins found issues. Prior to this we would not get any notification that problems existed unless something bigger happened, like a tool not being found. The result was that issues could exist but Travis CI would show successful builds.

statick_tool/exceptions.py
```diff
@@ -78,9 +78,11 @@ def filter_file_exceptions_early(self, package, file_list):
                 if exception["tools"] == 'all':
                     for pattern in exception["globs"]:
                         # Hack to avoid exceptions for everything on Travis CI.
-                        if pattern == '*/build/*':
-                            filename.lstrip('/home/travis/build/')
-                        if fnmatch.fnmatch(filename, pattern):
+                        fname = filename
```
tdenewiler: This fixes the `*/build/*` exception glob from being able to hide all the issues when building on Travis CI, which uses `/home/travis/build/` as the build directory. Identified in #65.

statick_tool/plugins/discovery/python_discovery_plugin.py
```diff
@@ -35,9 +35,11 @@ def scan(self, package, level, exceptions=None):
                     full_path = os.path.join(root, f)
                     output = subprocess.check_output(["file", full_path],
                                                      universal_newlines=True)
+                    # pylint: disable=unsupported-membership-test
```
tdenewiler: Would be nice to fix this rather than disable the warning.

statick_tool/plugins/tool/clang_tidy_tool_plugin.py
```diff
@@ -49,7 +49,7 @@ def scan(self, package, level):
             output = subprocess.check_output([clang_tidy_bin] + flags + files,
                                              stderr=subprocess.STDOUT,
                                              universal_newlines=True)
-            if "clang-diagnostic-error" in output:
+            if "clang-diagnostic-error" in output:  # pylint: disable=unsupported-membership-test
```
tdenewiler: Would be nice to fix this rather than disable the warning.

statick_tool/statick.py
```diff
@@ -6,6 +6,7 @@
 import logging
 import os
 
+from six import iteritems
```
tdenewiler: I could not figure out how to get Python 2/3 compatibility and fix the warning without bringing in this new module. I think it is a really common module to use in Python, so I'm okay with it.

PR #89, opened 2019-02-22
Regression fix: Add missing reporting plugins to install

creffett:
Minor whoopsie on my part - the reporting plugins weren't added to setup.py, so they're not getting installed (and so Statick isn't giving any output). This MR fixes that. Would appreciate a patch release after this is merged since this is a regression.


denewiler:
v0.2.1 released to PyPI with this fix.

PR #93, opened 2019-02-26
Add codecov support

creffett:
Let's see if this works.


creffett:
Appears to Just Work (TM)

PR #100, opened 2019-02-28
More tests, more coverage

creffett:
Work in progress, more commits to come. Adds more testing and increases code coverage.


creffett:
Marking as "ready for review" since Travis apparently won't run against draft requests, but not ready for review yet.

PR #102, opened 2019-03-18
Convert Findbugs to Spotbugs and make it run in Maven

creffett:
This removes the previous issues with Findbugs needing external commands to compile the Java files, but does mean that it will only work on Maven. Gradle and Ant integrations are also available, and I can add them in the future if needed.


statick_tool/plugins/tool/spotbugs_tool_plugin.py
```diff
@@ -0,0 +1,119 @@
...
+                    output = ex.output
+                    f.write(output)
+                    print("spotbugs failed! Returncode = {}".
+                          format(str(ex.returncode)))
+                    print("{}".format(ex.output))
+                    return None
+
+                except OSError as ex:
+                    print("Couldn't find maven! (%s)" % (ex))
+                    return None
```
denewiler: Earlier in this method you are returning `[]` instead of `None` when the `mvn` command is not found. Is there a reason to have a different return value?

Do you want to be checking again that this command exists? I think you should leave this extra check/error handling, since the first check may change in the future.

statick_tool/plugins/tool/spotbugs_tool_plugin.py
```diff
@@ -0,0 +1,119 @@
...
+                except subprocess.CalledProcessError as ex:
+                    output = ex.output
+                    f.write(output)
+                    print("spotbugs failed! Returncode = {}".
+                          format(str(ex.returncode)))
+                    print("{}".format(ex.output))
+                    return None
+
+                except OSError as ex:
+                    print("Couldn't find maven! (%s)" % (ex))
```
denewiler: I don't know that we are entirely consistent, but changing from `%` to `.format` seems like it would be closer to the rest of our print statements.

statick_tool/rsc/spotbugs-security.xml
```diff
@@ -1,5 +1,5 @@
 <FindBugsFilter>
```
denewiler: Should this (and the closing tag) be `SpotBugsFilter`?

statick_tool/plugins/tool/spotbugs_tool_plugin.py
creffett: I agree that they're inconsistent, but I'll bounce part of the question back to you - do you think a missing tool should be considered an error? I'm inclined to say "no" but I could see arguments both ways. I'd like to leave this check in just for safety's sake - I want to make sure that the subprocess call can't throw an unhandled exception, and I have seen times in the past where subprocess somehow couldn't find a tool which command_exists found (still trying to make a reproducible test case for that...)

statick_tool/rsc/spotbugs-security.xml
creffett: Surprisingly, no! Upstream kept it as `FindBugsFilter` for compatibility, and as far as I can tell they don't support `SpotBugsFilter` as a tag name.

statick_tool/plugins/tool/spotbugs_tool_plugin.py
denewiler: For considering a missing tool to be an error, I can see arguments both ways.

If you are expecting this tool to tell you when some constraints have been violated, and there is a specific tool to determine whether a set of constraints were violated, then a missing tool means you will not have any assurances that you are within desired constraints. That seems like something this tool should alert the user about. Right now, the only way I can think to create that alert is to return an error.

On the other hand, some users might want to use most of the tools for an existing level and are fine tracking upstream (so they don't have to create a custom config and merge upstream changes into it) but only installing the tools they care about. Or, some users might not be able to configure their tools for use in third party CI tools (such as Travis).

I'm not sure what all the use cases are for this tool. Right now I think it's only me and you :) I'm fine following your lead if you want to allow Statick to return success even when there are missing tools. In the future, I might go and add a "global" flag to Statick to allow the user to select how they want Statick to treat missing tools. Something like a `set_error_on_missing_tool` variable.

statick_tool/plugins/tool/spotbugs_tool_plugin.py
creffett: Will do.

PR #103, opened 2019-03-19
PEP3101 standardization

creffett:
Standardize on "".format instead of "%s" % ( ), add enforcement of that for the future.


denewiler:
Since #102 was merged there is a conflict with the `findbugs` tool plugin file. I'm not very familiar with editing in Github and don't want to make a hash of it. If you could merge master into your pep3101 branch I'd appreciate it.

creffett:
Will do.

PR #108, opened 2019-04-30
Error checking config

tdenewiler:


tdenewiler:
This has more than atomic changes. It should be split up into multiple PRs.

PR #136, opened 2019-08-14
Increasing test cases and coverage for cppcheck tool plugin.

tdenewiler:


denewiler:
Locally this gets all but lines 32 and 50. But in Travis there are a bunch more lines that are reported as not covered by unit tests. Merging, but I'll be looking at this again.

PR #157, opened 2019-11-17
Fixing bug in pyflakes output parsing. Pyflakes sometimes puts a colu…

### LINKED ISSUE #142: Support Python 3.8
According to [PEP-569](https://www.python.org/dev/peps/pep-0569/), Python 3.8 will be released on October 21, 2019. When that happens we should add it to the tox configuration.

tdenewiler:
…in Travis. This fixes #142.

### DIFF
--- a/statick_tool/plugins/tool/pyflakes_tool_plugin.py
+++ b/statick_tool/plugins/tool/pyflakes_tool_plugin.py
@@ -55,18 +55,48 @@ class PyflakesToolPlugin(ToolPlugin):
         issues = self.parse_output(total_output)
         return issues
 
-    def parse_output(self, total_output):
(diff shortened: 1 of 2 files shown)


tdenewiler:
An example of the type of file that would give multi-line output for pyflakes is:

```shell
$ echo "import filename" > pyflakes_test.py; echo "print 'Resizing %s' % input_filename" >> pyflakes_test.py
```

That script will give valid output:

```shell
$ python pyflakes_test.py 
Resizing
```

The output from pyflakes is 3 lines:

```shell
$ pyflakes pyflakes_test.py 
pyflakes_test.py:2:19: invalid syntax
print 'Resizing %s' % input_filename
                  ^
```

Issue #175, opened 2020-01-29
Remove IOError


tdenewiler:
There is a recognized confusion between `OSError` and `IOError`:

https://www.python.org/dev/peps/pep-3151/#confusing-set-of-os-related-exceptions

> In fact, it is hard to think of any situation where OSError should be caught but not IOError, or the reverse.

In Python 3.3 `IOError` was merged into `OSError`. Therefore, we should remove exception handling and unit tests for `IOError`.

https://docs.python.org/3/library/exceptions.html#OSError

PR #177, opened 2020-01-31
Replacing all usages of IOError with OSError. IOError was merged into…

tdenewiler:
… OSError in Python 3.3.


denewiler:
Fixes #175.

PR #179, opened 2020-01-31
Add support for Jenkins Warnings Next Generation plugin

axydes:
The Jenkins Warnings plugin has been deprecated and replaced with the Jenkins Warnings Next Generation Plugin. This change renames the write_file plugin to better represent what it does (write output compatible with the Jenkins Warnings plugin) and adds a new reporting plugin that supports the Jenkins Warnings Next Generation plugin.

https://wiki.jenkins.io/display/JENKINS/Warnings+Plugin
https://wiki.jenkins.io/display/JENKINS/Warnings+Next+Generation+Plugin


statick_tool/rsc/config.yaml
```diff
@@ -169,6 +184,9 @@ levels:
                        line-length: disable}}'"
 
   tex:
+    reporting:
+      print_to_console:
+      write_jenkins_warnings_ng:
```
tdenewiler: I don't think we want this plugin enabled for this level. No jobs I know of use this level with Jenkins.

statick_tool/rsc/config.yaml
```diff
@@ -103,6 +112,9 @@ levels:
         flags: "-checks='*,-cert-dcl50-cpp,-Wreserved-id-macro,-Weverything,-Wuser-defined-literals,-Wvexing-parse,-Misc-new-delete-overloads,-Cert-dcl59-cpp,-Wunsequenced,-analyzer-checker=cplusplus,-Wunevaluated-expression,-Wuniinitialized,-clang-analyzer-core.UndefinedBinaryOperatorResult,-Wdangling-initializer-list,-Wdelete-incomplete,-Wvarargs,-Winvalid-offsetof,-Wdynamic-class-memaccess,-wparentheses,-Wint-to-pointer-cast,-Wint-conversion,-Wincompatible-pointer-types,Clang-analyzer-cplusplus.NewDelete clang-analyzer-alpha.security.ArrayBoundV2,Clang-analyzer-cplusplus.NewDeleteLeaks-Wmismatched-new-delete,clang-analyzer-unix.MismatchedDeallocator,Clang-analyzer-unix.Malloc,Misc-non-copyable-objects,Cert-err52-cpp,-Wexceptions,Cert-err58-cpp,Cert-err60-cpp,Cert-err34-c,Cert-err34-c,-Clang-analyzer-alpha.cplusplus.VirtualCall,-Wdelete-non-virtual-dtor,-Wreorder,Cert-msc50-cpp,-Wreturn-type,-Winvalid-noreturn,Cert-msc30-c,Cert-flp30-c,Cert-env33-c'"
 
   self_check:
+    reporting:
+      print_to_console:
+      write_jenkins_warnings_ng:
```
tdenewiler: I don't think we want this plugin enabled for this level. No jobs I know of use this level with Jenkins.

statick_tool/plugins/reporting/write_jenkins_warnings_ng_reporting_plugin.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if not os.path.exists(output_dir):
+            os.mkdir(output_dir)
+        if not os.path.isdir(output_dir):
+            print("Unable to create output directory at {}!".format(
+                output_dir))
+            return None, False
+
+        output_file = os.path.join(output_dir,
+                                   package.name + "-" + level + ".statick")
```
axydes: Should I change the filename here? Maybe add a `.json` before the `.statick`? Otherwise if both the regular and ng version of the plugin are enabled (which if no reporting plugins are specified is the case), then they'll overwrite each other's files.

tdenewiler: That makes sense to me.

tox.ini
```diff
@@ -17,6 +17,7 @@ passenv = CI TRAVIS TRAVIS_*
 changedir = {toxinidir}/output-{envname}
 deps =
     codecov
+    deprecated
```
denewiler: Since this is used in an actual plugin and not in a unit test, I think this dependency should be removed from here and placed in `requirements.txt`.

axydes: If I remove this and try to run tox locally, it fails saying that it can't find deprecated. I'm not sure why as I have it installed.

PR #180, opened 2020-02-05
Clang format helpful error msg

tdenewiler:


statick_tool/plugins/tool/clang_format_tool_plugin.py
```diff
@@ -73,15 +74,28 @@ def scan(self, package, level):  # pylint: disable=too-many-locals, too-many-bra
...
+        except (IOError, OSError) as ex:
+            print("{}".format(exc_msg))
+            print("Error: {}".format(str(ex.strerror)))
+            if self.plugin_context.args.clang_format_raise_exception:
+                return None
+            return []
+
+        except subprocess.CalledProcessError as ex:
+            print("{} Returncode = {}".format(exc_msg, str(ex.returncode)))
+            print("Error: {}".format(ex.output))
```
xydesa: I think the `ex.output` string is the same as `exc_msg` in this case, which leads to a duplicate of that message being printed. That should be rectified. Maybe `exc_msg` is not needed in the first line?

PR #202, opened 2020-03-13
Mypy type hints

tdenewiler:


xydesa:
I'm not sure why, but when I run this branch I get the following error (that does not occur while running the pip installed version):
Can't find specified tool plugin eslint!

tdenewiler:
My guess as to why the `eslint` tool is not found is that the `--user-paths` argument needs to be updated to point at a location where that plugin is located. One way that worked for me was to append the system location to the user paths when running the local version of Statick. For example, instead of

    ./statick my_pkg --user-paths my_statick_config

use

  ./statick my_pkg --user-paths my_statick_config,/usr/local/lib/python3.6/dist-packages/statick_tool/

If you have the statick-web repo (that contains the eslint tool plugin) cloned locally, you could point your local statick at that repository instead of the system path.

xydesa:
Yup, good call. I forgot about that. With that modification this branch is working as normal.

Issue #204, opened 2020-03-18
Statick broken on Ubuntu 16.04-based distros.


kogut:
The current python3 version for Ubuntu 16.04 is 3.5.x:
https://packages.ubuntu.com/search?keywords=python3

But statick recently dropped support for python3 3.5:

https://github.com/sscpac/statick/commit/abd8a276023036f4d0e31114e00f963ba6be62ff#diff-2eeaed663bd0d25b7e608891384b7298L44

And running it on 16.04 (Mint 18) gives me this error for both `statick` and `statick_ws`:
```

Traceback (most recent call last):
  File "/usr/local/bin/statick_ws", line 9, in <module>
    from statick_tool.args import Args
  File "/usr/local/lib/python3.5/dist-packages/statick_tool/args.py", line 34
    user_paths: List[str] = []

              ^
SyntaxError: invalid syntax

```

python 3.5 end-of-life is scheduled for 20200913:

https://devguide.python.org/#status-of-python-branches

Just making sure this is an intentional decision to drop support for running statick on 16.04-based OSes.

tdenewiler:
Sorry, that was not intentional. I have access to a 16.04 machine and I'll work on fixing this. I'll make a fix and new release as soon as possible.

kogut:
The issue is fixed.

PR #205, opened 2020-03-19
Fixing bug in application of type hints that caused Statick with Pyth…

tdenewiler:
…on 3.5 to stop working.


denewiler:
Fixes #204

PR #207, opened 2020-03-21
Code formatting with black

tdenewiler:
Added CI checks to ensure future changes follow black formatting.


denewiler:
@axydes, if you have time to review I'd appreciate it.

denewiler:
Note that black is not part of the self_check level or part of the previously existing Actions file because black is not available for Python 3.5 or earlier. Therefore, I created a separate Actions file that runs on Python 3.6 and later to check for inconsistencies with black formatting.

PR #225, opened 2020-04-27
Standardize exit codes between statick and statick_ws. Only return no…

xydesa:
…n-success if --check is passed and issues were found.


denewiler:
Some of the motivation for this change is to follow Unix exit status codes: https://tldp.org/LDP/abs/html/exitcodes.html.

Another reason for the change is to allow statick output to go to stderr, then be piped to a separate process.

PR #249, opened 2020-08-11
Ubuntu 20.04 fixes for CMake discovery plugin

tdenewiler:
Fixes to CMake flags and template to get CMake discovery to work on Ubuntu 20.04. This was blocking any tools that depend on CMake or Make running.

### DIFF
--- a/statick_tool/plugins/discovery/cmake_discovery_plugin.py
+++ b/statick_tool/plugins/discovery/cmake_discovery_plugin.py
@@ -52,6 +52,7 @@ class CMakeDiscoveryPlugin(DiscoveryPlugin):
             "cmake",
             ".",
             "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
+            "-DBUILD_GMOCK=ON",
             "-DBUILD_GTEST=OFF",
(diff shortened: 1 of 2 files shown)


statick_tool/plugins/discovery/cmake_discovery_plugin.py
```diff
@@ -52,6 +52,7 @@ def scan(
             "cmake",
             ".",
             "-DCMAKE_BUILD_TYPE=RelWithDebInfo",
+            "-DBUILD_GMOCK=ON",
```
xydesa, line 55: not as part of this fix, but in the future it'd be good to extract these options into the statick config file.

xydesa:
It'd be nice if 20.04 were tested via github actions just like 16.04 and 18.04

statick_tool/plugins/discovery/cmake_discovery_plugin.py
tdenewiler: I agree. I created #250 to remind me.

tdenewiler:
Reply to https://github.com/sscpac/statick/pull/249#issuecomment-672267516:

I made #251 to track adding the `ubuntu-20.04` tag. It should be pretty easy, unless tests fail...

PR #264, opened 2020-09-29
Support alternate clang-format configuration file name. Allow the fil…

tdenewiler:
…e to be hidden in home directory.

### DIFF
--- a/statick_tool/plugins/tool/clang_format_tool_plugin.py
+++ b/statick_tool/plugins/tool/clang_format_tool_plugin.py
@@ -137,9 +137,14 @@ class ClangFormatToolPlugin(ToolPlugin):
             default_file_name = "_clang-format"
             format_file_name = self.plugin_context.resources.get_file(default_file_name)
             exc_msg = (
-                "_clang-format style is not correct. There is one located in {}. "
-                "Put this file in your home directory.".format(format_file_name)
+                "_clang-format or .clang-format style is not correct. There is "
+                "one located in {}. Put this file in your home directory.".format(
+                    format_file_name
+                )
             )
+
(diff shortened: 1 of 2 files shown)


statick_tool/plugins/tool/clang_format_tool_plugin.py
```diff
@@ -137,10 +137,15 @@ def check_configuration(self, clang_format_bin: str) -> Optional[bool]:
...
-                "_clang-format style is not correct. There is one located in {}. "
-                "Put this file in your home directory.".format(format_file_name)
+                "_clang-format or .clang-format style is not correct. There is "
+                "one located in {}. Put this file in your home directory.".format(
+                    format_file_name
+                )
             )
 
+            if not os.path.isfile(os.path.expanduser("~/" + default_file_name)):
+                default_file_name = ".clang-format"
```
xydesa, line 147: I think we should then do an 'isfile' test on this version, and throw an exception stating that the format file is missing

denewiler: The way it is here will not change any of the current behaviors. The exception message will be printed if neither file can be opened and the tool will return that it was not successful, but Statick will continue to run the next set of tools. If `~/_clang-format` does not exist then we switch to `~/.clang-format`. If `~/.clang-format` does not exist then it will be caught as an `IOError, OSError` when trying to open the file.

xydesa: Oh yea, I forgot about the code lower down catching that exception. I do think the message printed out might be slightly misleading for this situation though, it's not that the style of the file is incorrect so much as that the file is missing.

xydesa (approved):

PR #266, opened 2020-09-30
Allow globs for regex exceptions

tdenewiler:


xydesa:
Tested locally and it successfully allows a message on one file to be reported, but not a file that matches my glob.

xydesa (approved):

PR #270, opened 2020-10-16
Remove make tool dependence

tdenewiler:


xydesa (approved):
Tested locally and it works as desired. These plugins no longer require the make plugin enabled to run and they run fine.

PR #274, opened 2020-10-27
Add steps to generate and publish Sphinx documentation to GitHub Page…

tdenewiler:
…s on all new releases.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -62,3 +62,15 @@ jobs:
         uses: pypa/gh-action-pypi-publish@master
         with:
           password: ${{ secrets.PYPI_TOKEN }}
+
+      - name: Sphinx lint
+        uses: ammaraskar/sphinx-action@master
+        with:
+          docs-folder: "docs/"
+
+      - name: Publish documentation
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./docs/build/html
+          publish_branch: gh-pages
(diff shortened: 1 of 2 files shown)


tdenewiler:
I created a temporary commit that would run the Action to build and deploy documentation. Although that commit did not produce exactly the desired results it looks to me as if it is giving expected results. The Action for the temporary commit is at https://github.com/sscpac/statick/runs/1312149154?check_suite_focus=true. It made it all the way to the end, but failed when trying to commit the changes to the base repository with

```shell
Push the commit or tag
  /usr/bin/git push origin gh-pages
  remote: Permission to sscpac/statick.git denied to github-actions[bot].
  fatal: unable to access 'https://github.com/sscpac/statick.git/': The requested URL returned error: 403
  Error: Action failed with "The process '/usr/bin/git' failed with exit code 128"
```

This scenario involves trying to use `secrets.GITHUB_TOKEN` in a pull request, where the source branch is from a forked repository. The expected behavior for this scenario is described in [GitHub support](https://github.community/t/cant-push-to-forked-repository-on-the-original-repositorys-pull-request-with-github-token/16287).

> This is expected behavior, as mentioned here https://help.github.com/en/github/automating-your-workflow-with-github-actions
> /virtual-environments-for-github-actions#token-permissions 20 when running a pull_request form a fork the token can’t push
> contents to either repo.  This is to prevent melicious actors from using actions to poision upstrem or downstream repos.

We are seeing exactly what GitHub tells us to expect. I'm ready to cross my fingers that if this is merged it will cause documentation to build and be published when the next release is created. I can't think of a better way to test this. If it is broken, the worst case is documentation will not be built. But that is where we are without this change.

xydesa (approved):
Changes look good to me, and your explanation about the secrets also makes sense. I'm good with this.

Issue #275, opened 2020-10-30
CMake Discovery Plugin fails to find FastRTPS in a ROS2 package.


xydesa:
Replication steps:
```bash
git clone https://github.com/Liquid-ai/Plankton.git
cd Plankton
statick uuv_gazebo_worlds
```

Error:
```bash
Running cmake discovery plugin...
  Found cmake package /home/axydes/code/github/Plankton/uuv_gazebo_worlds/CMakeLists.txt
Problem running CMake! Returncode = 1
From /home/axydes/code/github/Plankton, running ['cmake', '.', '-DCMAKE_BUILD_TYPE=RelWithDebInfo', '-DBUILD_GMOCK=ON', '-DBUILD_GTEST=OFF', '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON', '-DCATKIN_ENABLE_TESTING=OFF', '-DCATKIN_SKIP_TESTING=ON', '-DINPUT_DIR=/home/axydes/code/github/Plankton/uuv_gazebo_worlds', '-DSTATICK_EXTRA_GCC_FLAGS=-Wcast-align -Wchar-subscripts -Wformat-security -Wformat -Wimplicit-int -Wincompatible-pointer-types -Wint-conversion -Wint-to-pointer-cast -Wparentheses -Winvalid-offsetof -Wsequence-point -Wdelete-incomplete -Wdelete-non-virtual-dtor -Winvalid-offsetof -Wreorder -Wreturn-type -Wuninitialized -Wunused-parameter -Wunreachable-code -Wunused-value -Wvarargs -Wtrigraphs -Wvarargs -Wformat -Wcast-qual']
-- STATICK FLAGS: -Wcast-align -Wchar-subscripts -Wformat-security -Wformat -Wimplicit-int -Wincompatible-pointer-types -Wint-conversion -Wint-to-pointer-cast -Wparentheses -Winvalid-offsetof -Wsequence-point -Wdelete-incomplete -Wdelete-non-virtual-dtor -Winvalid-offsetof -Wreorder -Wreturn-type -Wuninitialized -Wunused-parameter -Wunreachable-code -Wunused-value -Wvarargs -Wtrigraphs -Wvarargs -Wformat -Wcast-qual
-- PROJECT: [NAME:uuv_gazebo_worlds][SRC_DIR:/home/axydes/code/github/Plankton/uuv_gazebo_worlds][BIN_DIR:/home/axydes/code/github/Plankton]
-- HEADERS: 
[...]

PR #278, opened 2020-11-16
Add new ROS discovery plugin to replace the catkin discovery plugin. …

tdenewiler:
…The catkin discovery plugin is now marked as deprecated since the ROS one is more general. Add feature that allows discovery plugins to depend on other discovery plugins, necessary for getting additional CMake flags from ROS plugin to the CMake plugin.


xydesa (approved):
Tested locally and it works with both ROS 1 and ROS 2 packages. In addition, this fixes #275

PR #287, opened 2020-12-08
Swap order of checks in cpplint for package information. This will ca…

tdenewiler:
…use saner output for users when there are no C/C++ files found.

### DIFF
--- a/statick_tool/plugins/tool/cpplint_tool_plugin.py
+++ b/statick_tool/plugins/tool/cpplint_tool_plugin.py
@@ -18,12 +18,12 @@ class CpplintToolPlugin(ToolPlugin):
 
     def scan(self, package: Package, level: str) -> Optional[List[Issue]]:
         """Run tool and gather output."""
+        if "make_targets" not in package and "headers" not in package:
+            return []
+
         if "cpplint" not in package:
             print("  cpplint not found!")
             return None
-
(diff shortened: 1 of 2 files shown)


statick_tool/plugins/tool/cpplint_tool_plugin.py
```diff
@@ -18,13 +18,13 @@ def get_name(self) -> str:
 
     def scan(self, package: Package, level: str) -> Optional[List[Issue]]:
         """Run tool and gather output."""
+        if "make_targets" not in package and "headers" not in package:
```
xydesa, line 21: what about when `make_targets` or `headers` exist but are empty lists?

tdenewiler: I added an explicit check in the tool plugin that causes that method to return an empty list of issues as part of a8ec388. Prior to that, the plugin would also return an empty list of issues, and there was a unit test for that exact condition. I prefer to have it handled explicitly though, so that the intention is clear in the code. The previous way relied on unit tests to show the intention, but I could see that being non-obvious if a future developer were to go in and modify the tool plugin.

xydesa: Tested, it works now.

xydesa (approved):

PR #288, opened 2020-12-08
Add support to ROS discovery plugin for ROS2 Python-only packages tha…

tdenewiler:
…t do not contain CMakeLists.txt.


xydesa (commented):
the statick_ws tool also needs to be updated: https://github.com/sscpac/statick/blob/master/statick_ws#L54

xydesa (approved):
Tested locally, it detects a python-only ROS 2 package successfully.

PR #297, opened 2020-12-11
Add support to ROS discovery plugin for ROS2 Python-only packages tha…

tdenewiler:
…t do not contain CMakeLists.txt. Second attempt that fixed some bugs from first attempt.


xydesa (approved):
Tested locally, it finds the python-only ROS 2 packages in my workspace. Changes look good too. Thanks!

PR #302, opened 2020-12-11
Fix bug in lizard tool plugin where the directory to scan was not set…

tdenewiler:
… properly. Thanks @jhdcs for finding the bug.


jhdcs (approved):
This appears to be working correctly for me.

denewiler (approved):

PR #304, opened 2020-12-11
Ignoring any subdirectories in statick_ws if the current directory co…

xydesa:
…ntains CATKIN_IGNORE or AMENT_IGNORE.


denewiler (approved):

denewiler:
This worked for multiple test workspaces I have, including when I pointed `statick_ws` at the top-level of the workspaces (as opposed to the recommended way of pointing at the source directory). Duplicate packages were not discovered by `statick_ws` when pointed at the top-level workspace directory.

Looks good.

PR #305, opened 2020-12-11
Convert `lizard` tool plugin to use Python API

### LINKED ISSUE #298: Lizard does not accept flags from config
It appears that the `lizard` plugin doesn't use any user-specified flags from the `config.yaml` file. Checking the code at [lizard_tool_plugin.py](https://github.com/sscpac/statick/blob/master/statick_tool/plugins/tool/lizard_tool_plugin.py), the subprocess call in `scan` shows that user flags aren't apparently implemented...

### LINKED ISSUE #299: Investigate using lizard as Python module
A tool plugin for lizard already exists. That plugin uses `subprocess` to run the actual tool and get the output. However, there is a Python module available for lizard. We should investigate using the Python module instead of `subprocess`.

https://github.com/terryyin/lizard#using-lizard-as-python-module

jhdcs:
This primarily fixes #299, but _also_ fixes #298. Hopefully.

This is not the nicest looking code - Lizard basically wants to dump everything to `stdout` and then forget about it. The best I could figure out to do on the fly was to redirect that output using their own printing option, then reading that file back into Statick. Then, if the Statick user didn't want a file to begin with, I delete the lizard-generated file.

An improvement for a later date would probably be to capture the lizard output _before_ it gets to a file, so that Statick has full control over what gets written to the filesystem.

I'm not 100% confident on this code, so extra testing would be much appreciated.


jhdcs:
So it looks like there's something hardcoded in Statick that's expecting a supbprocess to happen. At least as far as `tox` is concerned. I'm not really familiar with that area of the code, so some insight would be lovely...

Work on this will probably resume after the weekend.

xydesa:
> So it looks like there's something hardcoded in Statick that's expecting a supbprocess to happen. At least as far as `tox` is concerned. I'm not really familiar with that area of the code, so some insight would be lovely...
> 
> Work on this will probably resume after the weekend.

There's a unit test that specifically tests the subprocess functionality, this may be causing your issue:
https://github.com/sscpac/statick/blob/master/tests/plugins/tool/lizard_tool_plugin/test_lizard_tool_plugin.py#L131

It would be good to update that unit test to test your changes once you're confident in how the new plugin is using lizard.

jhdcs:
I'm concerned that test coverage was reduced. It appears to be because there aren't tests checking what happens when different flags are passed to `lizard`, but I'm not finding any examples in the other tool plugins where we test tool flags. Am I blind?

tdenewiler:
@jhdcs, for the unit tests it looks like you are going to need to load a custom configuration file in order to set options to pass to the `lizard` module. The `spotbugs` unit tests use a custom configuration file. The custom configuration files are at:

https://github.com/sscpac/statick/tree/master/tests/plugins/tool/spotbugs_tool_plugin/rsc

When setting up the tests there is an option to pass in a `custom_rsc_path`. See:

https://github.com/sscpac/statick/blob/master/tests/plugins/tool/spotbugs_tool_plugin/test_spotbugs_tool_plugin.py#L20

One of the unit tests then passes in the path to the custom configuration files.

https://github.com/sscpac/statick/blob/master/tests/plugins/tool/spotbugs_tool_plugin/test_spotbugs_tool_plugin.py#L138

I think this pattern is the way to get coverage of the new lines where options are passed in to the lizard tool.

jhdcs:
The more I poke at this, the more I wonder if preserving the options creating the decline in code coverage is worth it.

For the "schema.any_silent()" check at https://github.com/sscpac/statick/blob/1cc496c03d295d40dc59c52a86395719815a932a/statick_tool/plugins/tool/lizard_tool_plugin.py#L38 that only comes into play if you use the "-Edumpcomments" flag - which is practically useless for Statick, since all it does is just print out all the comments in the scanned files to `stdout`, and hides the issues from being discovered.

https://github.com/sscpac/statick/blob/1cc496c03d295d40dc59c52a86395719815a932a/statick_tool/plugins/tool/lizard_tool_plugin.py#L41 Could be potentially useful... If it wasn't for the fact that Statick is already taking care of this.

https://github.com/sscpac/statick/blob/1cc496c03d295d40dc59c52a86395719815a932a/statick_tool/plugins/tool/lizard_tool_plugin.py#L44 Is in much the same boat. Statick is already providing the output file. This embedded `lizard` doesn't have to worry about that.

I'm in favor of removing these options from the embedded `lizard` - but if I do, how should it be documented that these options are no longer supported?

tdenewiler:
@jhdcs, I am in favor of removing support for the options you listed. I don't think it's critical to document that those options are not respected, but if you want to document it I think the top-level docstring of the `lizard_tool_plugin.py` file would be a good place to do so.

Hopefully later today I can test out your branch on some of my workspaces. But, this is looking pretty close to finished to me. If you need any help getting complete code coverage just drop me a line.

jhdcs:
I modified the code a little bit to improve testability (hopefully). Barring what I've just done, I'm not sure what should be done.

It looks like the code coverage calculator is flagging a few files that weren't modified. Is that normal?

tdenewiler:
@jhdcs, the error with tox was happening due to unit tests that mock exceptions from the subprocess module. We use mock so that our unit tests can get to branches in the code where exception handling occurs -- it is easier than hand crafting files that will cause exceptions. Now that you have switched from subprocess to importing the lizard module there are no exceptions thrown in the lizard tool plugin by calling `subprocess.check_output`. I sent you a PR to fix the tox output. There are still some lines that are not covered by unit tests though. Let me know if you need help with that part. Note that I did not really test or review the changes yet, but I plan to do that by early next week.

tdenewiler:
@jhdcs, I sent you a PR to do a bit of minor cleanup. The branch I sent you appears to work as expected on multiple test packages and workspaces that I have locally.

I removed the try/except block because I don't see anything that raises an exception that we can catch in the lizard module. The `io.StringIO` module can throw an exception (`ValueError` and `io.UnsupportedOperation`) but I could not figure out a way to test for that. The main reason they are difficult to test is that it enters C code instead of Python code and `mock` has a difficult time changing that type of output. I'll try to come up with a solution for that later.

https://docs.python.org/3/library/io.html#io.IOBase
https://blog.xelnor.net/python-mocking-datetime/

The code coverage often shows inconsistent results on pull requests. The primary cause is that tests are occurring on Windows, Mac, and Linux. When the Windows or Mac tests finish first and those results are uploaded to Codecov, Codecov pushes out the results. However, many tests are not run on Windows, mostly due to not installing tool binaries in the CI environment. After all test environments are completed and results uploaded, Codecov is smart enough to merge the results files together. But Codecov does not update the results pushed to the PR.

I run tests locally using `tox -e py38` on a Linux Mint 20 machine (derivative of Ubuntu 20.04). On that system I expect to see a single line not covered by unit test (in the `discovery_plugin` module). During CI runs that line is covered by the tests on Windows.

jhdcs:
Thanks for the cleanup! I've merged it into the PR branch.

denewiler (approved):

denewiler:
Thanks @jhdcs!

PR #308, opened 2020-12-22
Discovery walk once

### LINKED ISSUE #282: Discovery should walk once
Currently, each individual discovery plugin is responsible for walking the entire package path to search for files to scan. When there are a large number of files and a large number of discovery plugins enabled, the discovery phase can take a long time to complete. We should create a new method in the `DiscoveryPlugin` class that walks the package path and stores all of the information about each file in a dictionary, including filename, full path, and output of the `file` command. Then, each discovery plugin can search through that dictionary to discover the files that match their pattern.

This could break any third party discovery plugins. I am not aware of any discovery plugins where I am not a contributor. If this is done, it would be nice that we do not require (at first) that plugins use the master dictionary. The first way I can think of to handle that is to keep the current name of `scan` that is called by the main Statick program for each discovery plugin. However, in each of the discovery plugins under our control we can have the `scan` method call search the master dictionary. The `scan` method would still use that to fill in `package[file_type_src]`, so the interfaces would remain the same.

When working on this, it would be nice to identify a benchmark repository. This repository should have a large size (I'm thinking ~2GB), and have multiple file types that would exercise our discovery plugins. We could then create a temporary level that has all the discovery plugins enabled, but none of the tool or reporting plugins. Then we could run Statick and time how long it takes to complete.

```shell
time statick large_repository --profile discovery_only.yaml
```

xydesa:
Fixes #282 

This should not break or change any un-modified discovery plugins.

I timed both the master branch and this issue branch against the nodejs repo: https://github.com/nodejs/node.git
The issue branch finds 1 more perl file than the master branch, which I think is a file that the master branch was just missing:
node/deps/openssl/openssl/external/perl/Text-Template-1.46/Makefile.PL

The timing on the issue branch was done using branches for the other statick repos as well (statick-md, statick-web, statick-tex). Once this branch is approved, I'll open PRs for the other repositories.

master:
real	15m3.657s
user	10m31.563s
sys	5m29.671s

issue:
real	1m44.862s
user	1m17.770s
sys	0m33.676s

Here is all the actual output:
[issue-output.log](https://github.com/sscpac/statick/files/5730996/issue-output.log)
[master-output.log](https://github.com/sscpac/statick/files/5730997/master-output.log)
[new-files.log](https://github.com/sscpac/statick/files/5730998/new-files.log)
[old-files.log](https://github.com/sscpac/statick/files/5730999/old-files.log)

### DIFF
--- a/statick_tool/discovery_plugin.py
+++ b/statick_tool/discovery_plugin.py
@@ -57,7 +57,20 @@ class DiscoveryPlugin(IPlugin):  # type: ignore
         package._Package__walked = True
 
     def get_file_cmd_output(self, full_path: str) -> str:
-        """Run the file command (if it exists) on the supplied path."""
+        """
+        Run the file command (if it exists) on the supplied path.
+
+        The output from the file command is converted to lowercase.
+        There are two recommended ways to check it:
+        1. When searching for a single string just use the python "in" operator:
+
+            if "search string" in fild_dict["file_cmd_out"]:
+
+        2. When searching for multiple different strings, use the `any()` function:
+
+            expected_output = ("output_1", "output_2")
+            if any(item in file_dict["file_cmd_out"] for item in expected_output):
+        """
         if not self.file_command_exists():
             return ""
 
--- a/statick_tool/discovery_plugin.py
+++ b/statick_tool/discovery_plugin.py
@@ -62,7 +62,9 @@ class DiscoveryPlugin(IPlugin):  # type: ignore
                 ["file", full_path], universal_newlines=True
             )  # type: str
             return output
-        except subprocess.CalledProcessError:
+        except subprocess.CalledProcessError as ex:
+            print("Failed to run 'file' command. Returncode = {}".format(ex.returncode))
+            print("Exception output: {}".format(ex.output))
             return ""
 
     def set_plugin_context(self, plugin_context: Union[None, PluginContext]) -> None:
--- a/statick_tool/discovery_plugin.py
+++ b/statick_tool/discovery_plugin.py
@@ -37,16 +37,9 @@ class DiscoveryPlugin(IPlugin):  # type: ignore
         use it to filter which files the plugin detects.
         """
 
-    def scan_once(
-        self, package: Package, level: str, exceptions: Optional[Exceptions] = None
-    ) -> None:
-        """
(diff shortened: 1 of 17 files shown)


xydesa:
Let me know if you want me to add the `discovery_only` profile to this PR.

statick_tool/discovery_plugin.py
```diff
@@ -36,6 +37,40 @@ def scan(
...
+        """Run the file command (if it exists) on the supplied path."""
+        if not self.file_command_exists():
+            return ""
+
+        try:
+            output = subprocess.check_output(
+                ["file", full_path], universal_newlines=True
+            )  # type: str
+            return output
+        except subprocess.CalledProcessError as ex:
```
xydesa, line 69: I'm not completely sure how this is being tested, but the coverage reports say it is. I also didn't find any unit tests specifically for this file. Do you have any suggestions on whether I should add a new unit test specifically for this file?

tdenewiler:
The `disocvery_only` profile sounds like a nice addition for testing. Please add it.

tdenewiler:
I think we should add explicit unit tests for the Discovery plugin. It would be helpful to run only those unit tests and see that the newly added methods are fully covered.

tdenewiler:
I ran a comparison of Statick with the `discovery_only` profile on both the PR branch and master. On master I simply copied the `config.yaml` and `discovery_only.yaml` files from the PR branch into the `rsc` directory. The package I used for testing was the Linux kernel. The PR branch found 1 extra Perl file. The rest of the discovery plugins found the same number of files for both initial and filtered passes.

For this particular package/repository (Linux kernel) the PR branch improves discovery by about 3x. That is a significant improvement.

**Master Branch**

```
Statick exiting with success.

real	16m17.807s
user	10m27.757s
sys	6m50.729s
```

**PR Branch**

```
Statick exiting with success.

real	5m27.114s
user	3m11.956s
sys	2m33.926s
```

I ran the same test on a small ROS package and got similar 3x improvements.

**Master Branch**

```
Statick exiting with success.

real	0m15.055s
user	0m10.147s
sys	0m6.521s
```

**PR Branch**

```
Statick exiting with success.

real	0m5.035s
user	0m3.284s
sys	0m1.843s
```

denewiler (approved):

PR #311, opened 2020-12-23
Multiprocess statick_ws

xydesa:
This works, and runs faster than the master branch. However I cannot figure out how (or indeed if it's even possible) to make the package count incrementally increase (i.e. to maintain a count of where in the list of packages statick is).

On this workspace I see the below difference in timings: https://github.com/Field-Robotics-Lab/dave/blob/master/repos/dave_sim.repos
`time statick_ws src --output-directory /tmp/statick-output`
master:
Statick exiting with success.

real	0m21.468s
user	0m17.024s
sys	0m5.003s

issue:
Statick exiting with success.

real	0m4.316s
user	0m18.927s
sys	0m5.157s

### DIFF
--- a/statick_ws
+++ b/statick_ws
@@ -3,6 +3,7 @@
 
 from __future__ import print_function
 
+import argparse
 import multiprocessing
 import os
 import sys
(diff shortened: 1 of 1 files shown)


tdenewiler:
Here are timings I got for a ROS workspace.

*Issue*

```
Statick exiting with success.

real	39m42.741s
user	83m53.526s
sys	3m29.457s
```

*Installed via pip 0.4.11*

```
Statick exiting with success.

real	74m42.130s
user	72m18.123s
sys	2m39.055s
```

I also noticed via `htop` that a lot more CPU cores were utilized with the issue branch.

I was able to test that appending `--max_procs 1` defaulted to the previous behavior of scanning a single package at a time.

denewiler (approved):

PR #319, opened 2021-01-11
Remove trailing colon from filename when adding issues for black tool.

tdenewiler:


tdenewiler:
I found this bug when running Statick against [errbot](https://github.com/errbotio/errbot). For the Statick configuration I used one from [make_movie](https://github.com/tdenewiler/make_movie/tree/master/statick_config/rsc). But any level that has the `black` tool enabled will show the bug.

```shell
statick errbot --output-directory /tmp/statick-output --user-paths make_movie/statick_config --force-tool-list black
```

With the `master` branch there is an error when running black against `README.rst`. Note the trailing colon on the filename.

```shell
  File "/home/thomas/.local/bin/statick", line 46, in <module>
    main()
  File "/home/thomas/.local/bin/statick", line 28, in main
  File "/home/thomas/.local/lib/python3.8/site-packages/statick_tool/statick.py", line 346, in run
    issues = self.exceptions.filter_issues(package, issues)
  File "/home/thomas/.local/lib/python3.8/site-packages/statick_tool/exceptions.py", line 221, in filter_issues
    issues = self.filter_nolint(issues)
  File "/home/thomas/.local/lib/python3.8/site-packages/statick_tool/exceptions.py", line 202, in filter_nolint
    lines = open(issue.filename, "r+", encoding="utf-8").readlines()
FileNotFoundError: [Errno 2] No such file or directory: '/home/thomas/src/errbot/README.rst:'
```

As an aside, this is the first time I learned that the Unix `file` command says that `.rst` files are Python files. Perhaps we should check for that extension and add it to a new file type so that documentation linters can scan the files. There is little value in running `pylint` or `black` against those files.

xydesa (commented):
Tested locally, and the change works.
Perhaps we should add some parsing to this plugin to detect an internal error from `Black` like the below, and add it to the issues instead of the weird text that does get added:
```
Black encountered internal error
error: cannot format /home/axydes/code/github/errbot/README.rst: Cannot parse: 1:3: .. image:: https://errbot.readthedocs.org/en/latest/_static/errbot.png
Oh no! 💥 💔 💥
1 file would fail to reformat.

Black encountered internal error
error: cannot format /home/axydes/code/github/errbot/errbot/templates/new_plugin.py.tmpl: Cannot parse: 4:6: class {{ class_name }}(BotPlugin):
Oh no! 💥 💔 💥
1 file would fail to reformat.

black tool plugin done.
```

```
Tool black: 3 unique issues
  /home/axydes/code/github/errbot/README.rst:0: black:format: .. image: [3]
  /home/axydes/code/github/errbot/setup.py:0: black:format: would reformat [3]
  /home/axydes/code/github/errbot/errbot/templates/new_plugin.py.tmpl:0: black:format: 4:6 [3]
3 total unique issues
```

tdenewiler:
@xydesa I agree that the output does not look good. I'm open to suggestions on improvements. One thing we should do for sure is not print anything if the `show-tool-output` flag is set to False. Perhaps we could make the default output of the tool plugin be "Black encountered internal error while scanning <filename>"? Then the full output would show if `show-tool-output` is set to True. The docformatter output is similarly overwhelming.

The black plugin does detect those internal errors.

https://github.com/sscpac/statick/blob/master/statick_tool/plugins/tool/black_tool_plugin.py#L35

See the notes for the `--check` flag in black for where the return code of 123 comes from..

https://black.readthedocs.io/en/stable/installation_and_usage.html#command-line-options

During parsing of the errors the plugin differentiates between whether an internal error was found or a file would be reformatted. Both types of issues are reported at the same warning level.

xydesa (approved):
Works when tested locally. Will be opening a separate pr to improve parsing black's internal errors. And I'm fine with a separate PR to fix the tool output as well.

PR #325, opened 2021-01-12
Check for a valid plugin context before accessing plugin context vari…

tdenewiler:
…ables related to showing tool output.

### DIFF
--- a/statick_tool/plugins/tool/cccc_tool_plugin.py
+++ b/statick_tool/plugins/tool/cccc_tool_plugin.py
@@ -77,7 +77,7 @@ class CCCCToolPlugin(ToolPlugin):
                 print("Couldn't find cccc executable! ({})".format(ex))
                 return None
 
-            if self.plugin_context.args.show_tool_output:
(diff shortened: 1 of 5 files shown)


statick_tool/plugins/tool/cccc_tool_plugin.py
```diff
@@ -77,7 +77,7 @@ def scan(  # pylint: disable=too-many-branches
                 print("Couldn't find cccc executable! ({})".format(ex))
                 return None
 
-            if self.plugin_context.args.show_tool_output:
+            if self.plugin_context and self.plugin_context.args.show_tool_output:
```
xydesa, line 80: this tool checks for self.plugin_context is None at the top of this function. is this change still necessary?

tdenewiler: Statick can still crash if `self.plugin_context` is empty. I've seen that happen a few times in the past.

xydesa (approved):

PR #326, opened 2021-01-12
Check for a valid plugin context before accessing plugin context vari…

tdenewiler:
…ables related to the existence of an output directory.


xydesa (approved):

xydesa (changes requested):
looks like unit test coverage goes down a bit.

Update: after a closer look, it seems to be missing the windows specific lines. maybe the report wasn't uploaded from windows correctly?

tdenewiler:
I saw the same thing as you on codecov. I re-ran the PR action and the codecov report was updated to show full line coverage. Looks like that was a false positive, worth checking out.

xydesa (approved):

PR #335, opened 2021-01-22
Automatically creating the output directory if it does not exist.

xydesa:
Signed-off-by: Alexander Xydes <alexander.xydes@navy.mil>


denewiler (approved):
I tested locally as well and the `--output-directory` was created without causing an error. Nice update.

PR #336, opened 2021-01-25
Logging memory handler

### LINKED ISSUE #331: Running statick on a workspace at INFO log level results in logging statement of different packages interleaved among each other
When the print statement was used, the [`scan_package`](https://github.com/sscpac/statick/blob/master/statick_tool/statick.py#L574) function captured the stdout and stderr while running statick on a single package and then printed everything from that package at once.

I don't know if this is possible with the logging module, but we should look into it or something like it. What's there now is not working.

xydesa:
Fixes #331

### DIFF
--- a/statick_tool/statick.py
+++ b/statick_tool/statick.py
@@ -597,20 +597,20 @@ class Statick:
         )
         issues, dummy = self.run(package.path, parsed_args)
         if issues is not None:
+            sys.stdout = old_stdout
+            sys.stderr = old_stderr
+            logging.info(sio.getvalue())
             logging.info(
                 "-- Done scanning package %s (%d of %d) --",
                 package.name,
                 count,
                 num_packages,
             )
-            sys.stdout = old_stdout
-            sys.stderr = old_stderr
-            logging.info(sio.getvalue())
         else:
-            logging.error("Failed to run statick on package %s!", package.name)
             sys.stdout = old_stdout
             sys.stderr = old_stderr
             logging.info(sio.getvalue())
+            logging.error("Failed to run statick on package %s!", package.name)
         if old_handler is not None:
             handler.flush()
             logger.removeHandler(handler)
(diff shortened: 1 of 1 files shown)


statick_tool/statick.py
```diff
@@ -579,30 +580,45 @@ def scan_package(
         num_packages: int,
     ) -> Optional[Dict[str, List[Issue]]]:
         """Scan each package in a separate process while buffering output."""
+        logger = logging.getLogger()
+        old_handler = None
+        if logger.handlers[0]:
+            old_handler = logger.handlers[0]
+            handler = MemoryHandler(10000, flushLevel=logging.ERROR, target=old_handler)
```
xydesa, line 587: the `10000` is how many log statements it will buffer before automatically flushing to the target handler. I chose a number that I thought would not be easily reached by a single call to `run`.
the `flushLevel` indicates that if a log statement comes it at the `ERROR` level the handler will immediately flush it's buffer. it seemed to me that for how we use the log levels, anything error and above should be immediately printed. if that should be changed we have two options:
1. up that to critical, which we don't currently use.
2. create our own subclass of the BufferingHandler that implements the `shouldFlush` function as we want
https://docs.python.org/3/library/logging.handlers.html#memoryhandler

tdenewiler: The `10000` number seems fine to me. That is per package, right? I have only occasionally seen that many issues for large workspaces, so I think it will be fine either way.

I like your approach to printing out `ERROR` and higher immediately. I think that is a great place to start. If we or other users find that a different approach should be taken then we can re-visit the other options you listed.

xydesa: Yes. It'll be per call to `scan_package`, which is called with a single package once per subprocess.

tdenewiler:
This worked for me on multiple workspaces.

[DAVE](https://github.com/Field-Robotics-Lab/dave/blob/master/repos/dave_sim.repos)
[Node Example](https://github.com/tdenewiler/node_example/blob/main/rosinstall/node_example.rosinstall)

denewiler (approved):

PR #337, opened 2021-01-26
Allow custom configuration levels to inherit from base levels.

tdenewiler:


statick_tool/config.py
```diff
@@ -15,18 +15,49 @@ class Config:
...
-            try:
-                self.config = yaml.safe_load(fname)
-            except (yaml.YAMLError, yaml.scanner.ScannerError) as ex:
-                raise ValueError(
-                    "{} is not a valid YAML file: {}".format(filename, ex)
-                ) from ex
+
+        self.config = self.get_config_from_file(base_file)
+
+        if user_file:
```
xydesa, line 26: shouldn't you check that the path exists here as well?

xydesa (approved):

Issue #344, opened 2021-02-10
PermissionError when executing filter_nolint


mlanting:
Occasionally when Statick is scanning a package, it will report errors from a dependency instead of the package itself. Usually this just means the output for one package might report issues from another package in our workspace. Recently, however, Statick jumped into a dependency in the /opt directory where everything is set to read-only and is throwing a PermissionError exception when attempting to open a source file.

In https://github.com/sscpac/statick/blob/master/statick_tool/exceptions.py, it looks like some of the functions are opening files in read-write mode when they only need to read them.

tdenewiler:
Can you give the `open-as-read-only` branch a try on your packages? That branch is in the tdenewiler fork.

https://github.com/tdenewiler/statick/tree/open-as-read-only

mlanting:
When I try to run `sudo python3 setup.py install` I get the error:
`error: can't copy 'statick_tool/rsc/plugin_mapping': doesn't exist or not a regular file`

I tried running it right from the cloned repo and it's giving me a bunch of errors about being unable to create output directories for the packages. I'm using basically the same command the docker is using, and the docker doesn't seem to have the issue of being unable to create output directories.

The Docker script uses something like:
`statick /home/admin/project/src --output-directory output -ws`

I'm using:
`statick/statick /home/admin/project/src --output-directory output -ws`
after cloning statick and switching to your branch.

I'm doing this inside of the docker container launched in interactive mode.

tdenewiler:
Sounds like you have done a lot of testing. I don't want you to have to spend more time on that. We merged what I think will be a fairly simple fix to the problem, and I'll make a new release now.

mlanting:
alright, i'll give it a try with the latest release. Thanks!

tdenewiler:
New release is published on PyPI for v0.5.2. Please let me know if you still have this issue.

mlanting:
I was able to complete a scan without seeing the issue using v0.5.2. Thanks for your help!

PR #346, opened 2021-02-10
In the exceptions module, open files in read-only mode when filtering…

tdenewiler:
… the lines for the NOLINT string.


tdenewiler:
Fixes #344.

xydesa (approved):
Tested locally by changing a package to read-only. This fix works.

PR #358, opened 2021-05-14
Remove testing support for Ubuntu 16.04 and Python 3.5. There is no g…

tdenewiler:
…uarantee Statick will work in those environments any longer.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,7 @@ jobs:
           sudo apt install uncrustify
 
       - name: Install clang tools (for unit tests)
-        if: matrix.os == 'ubuntu-18.04'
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -98,14 +98,6 @@ jobs:
           sudo npm install -g markdownlint-cli@0.21.0
           rm nodesource_setup.sh
 
-      # Black is only available on Python 3.6+, but we still want to test Python 3.5 since that is
```
xydesa, line 101: shouldn't line 82 above be for 18.04 or 20.04?

xydesa:
I think tox.ini needs to be updated too

xydesa (approved):

PR #362, opened 2021-05-17
Use f-strings for all print statements when variables are referenced. 

tdenewiler:


tdenewiler:
To find print statements to change I used this grep pattern after going into the `statick_tool`, `tests`, and `examples` directories. I changed any statements that referenced variables.

```sh
grep -rn "print(\""
```

xydesa (approved):
Tested locally and things still get printed correctly. LGTM.

PR #363, opened 2021-05-18
Switch type hints from comment style to inline style. 

tdenewiler:
I looked for comment style type hints by using the following grep search in `statick_tool`, `tests`, and `examples`.

```sh
grep -rn "# type:" | grep -v ignore
```


xydesa (approved):
Tested locally and it ran fine. LGTM.

PR #370, opened 2021-06-07
Adding groovy discovery plugin and npm-groovy-lint tool plugins.

xydesa:


docs/source/statick_tool.plugins.discovery.rst
```diff
@@ -30,6 +30,15 @@ statick_tool.plugins.discovery.cmake_discovery_plugin module
     :show-inheritance:
 
 
+statick_tool.plugins.discovery.groovy_discovery_plugin module
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
denewiler, line 34: I think the tildes should go to the end of the text on the line above. Sphinx-lint might catch that.

xydesa: Apparently sphinx lint did not catch that.

denewiler: https://github.com/sscpac/statick/runs/2766582228?check_suite_focus=true#step:15:153

It is showing up as a warning and not being elevated to visibility at this level. I'm not going to worry about that here. It could be a separate issue to find out why, and to discuss if we want to enforce that.

denewiler (approved):

PR #383, opened 2021-09-08
Workflow trigger 383

gregtkogut:
This pull request might be best tested in a private fork.   I couldn't think of a better way .

For two reasons - it's not easy to verify that the build is actually triggered at the scheduled time if not on the default branch of a repo.  

Also the button for manually triggering a workflow only appears on the default branch of a repo.   So to test these capabilities I changed the default branch of my personal fork to the issue branch.    That's not something advisable on a "production" repo like this one since it can break things, like any active PRs.

Attached here a screenshot of where the button to manually trigger a workflow run should appear.  This was on my personal fork, after having changed the default branch from 'main' to the issue branch.

manual_run

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,12 @@
 ---
 name: Statick
 
-on: [pull_request, push, workflow_dispatch]  # NOLINT
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * MON'
+  workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


gregtkogut:
...I selected Monday at 3AM for the run based on commentary with other Collaborators, but have no strong feelings about that particular time.

xydesa (approved):
These changes match what I found when I was looking into scheduling github workflow actions. LGTM.

denewiler (approved):
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

LGTM

PR #395, opened 2021-12-09
Iterate directly on dictionary instead of calling .keys().

tdenewiler:


tdenewiler:
This fixes pylint issue that showed up in the self test of the most recent weekly scheduled action.

xydesa:
Tested locally, still works.

xydesa (approved):

PR #396, opened 2021-12-09
Allow stand-alone Python packages to be discovered as workspace packages for parallel scans.

tdenewiler:


xydesa (approved):
This works for me in my local workspaces.

PR #397, opened 2021-12-10
Change all Python versionsn in workflow yaml files to be strings.

tdenewiler:


xydesa:
What prompted this change?

tdenewiler:
Yaml is crazy. I had at least one repo where `3.10` turned into `3.1`, as the trailing zero was removed. That failed a workflow because Python 3.1 did not exist as an option. The fix was to add double quotes around the values to make them strings. Some more information from Stack Overflow:

https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml

xydesa:
based on that link, it seems like you should use single quotes whenever possible, link in this case. since you don't need special characters. and in fact, the rest of the file uses single quotes for strings. so i think you should switch to single quotes.

xydesa (approved):

tdenewiler:
For more context on the motivation for these changes see #375, in particular https://dev.to/hugovk/the-python-3-1-problem-85g.

PR #399, opened 2022-01-04
Add list of known third party plugins to README.

tdenewiler:


README.md
```diff
@@ -721,6 +722,20 @@ black statick statick_tool tests
 docformatter -i --wrap-summaries 88 --wrap-descriptions 88 <file>
 ```
 
+## Third Party Plugins
+
+Known third party Statick plugins.
+
+| Plugin Name      | Repository Location                              |
+| :--------------- | :----------------------------------------------- |
+| statick-action   | <https://github.com/sscpac/statick-action>       |
```
xydesa, line 731: i don't think i would call this a plugin, that seems to imply that it will lint github action files to me. I'd add another section calling out that we have a github action. I think it makes sense to put that in an "Installation" section which I've seen other software do (like mega-linter: https://megalinter.github.io/latest/installation/)

README.md
```diff
@@ -721,6 +722,20 @@ black statick statick_tool tests
 docformatter -i --wrap-summaries 88 --wrap-descriptions 88 <file>
 ```
 
+## Third Party Plugins
```
xydesa, line 725: some of these plugins are not from third parties, maybe `External Plugins` would be more accurate?

xydesa (changes requested):

xydesa (approved):

PR #405, opened 2022-03-03
Allow a level to inherit configuration from multiple levels.

tdenewiler:


statick_tool/plugins/tool/xmllint_tool_plugin.py
```diff
@@ -21,9 +21,13 @@ def scan(self, package: Package, level: str) -> Optional[List[Issue]]:
         flags: List[str] = []
         flags += self.get_user_flags(level)
 
+        files: List[str] = []
```
xydesa, line 24: didn't these changes get merged in #403? how come they're still listed here

tests/config/rsc/user.yaml
```diff
@@ -1,6 +1,7 @@
 levels:
   custom:
-    inherits_from: "threshold"
+    inherits_from:
```
xydesa, line 3: what about testing inheriting from levels in both the base config file and the custom file?

statick_tool/plugins/tool/xmllint_tool_plugin.py
tdenewiler: I merged upstream (sscpac) main into my source branch and cleaned that up.

gregtkogut (approved):

tests/config/rsc/user.yaml
xydesa: that tests for multiple levels from one file, but not one level from the custom config and one level from the base statick config.

tdenewiler: [b7ee16e](https://github.com/sscpac/statick/pull/405/commits/b7ee16ec5c1d6e31a0c4f2bda83656cfac839111). This test uses a base configuration file and a user configuration file. The level we are testing inherits from levels in both base and user configuration files. I had to make the test `config.yaml` have reporting plugins to test that part, where the addition matches the main Statick configuration file.

xydesa (approved):

PR #412, opened 2022-04-02
CCCC tool plugin revamp

xydesa:
This change will likely result in more issues being found by this tool. Prepare accordingly.

### DIFF
--- a/statick_tool/plugins/tool/cccc_tool_plugin.py
+++ b/statick_tool/plugins/tool/cccc_tool_plugin.py
@@ -72,7 +72,7 @@ class CCCCToolPlugin(ToolPlugin):
 
             try:
                 subproc_args: List[str] = [cccc_bin] + temp_opts + [src]
-                logging.debug(' ' .join(subproc_args))
+                logging.debug(" ".join(subproc_args))
                 log_output: bytes = subprocess.check_output(
                     subproc_args, stderr=subprocess.STDOUT
                 )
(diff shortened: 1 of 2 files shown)


statick_tool/plugins/tool/cccc_tool_plugin.py
```diff
@@ -56,13 +58,21 @@ def scan(  # pylint: disable=too-many-branches
...
         else:
             return []
+        opts.append(" --lang=c++")
+
+        issues: List[Issue] = []
 
         for src in package["c_src"]:
+            tool_output_dir: str = ".cccc-" + Path(src).name
+            temp_opts = opts
+            temp_opts.append("--outdir=" + tool_output_dir)
```
denewiler, line 71: I tested locally and don't see the need for the temporary variable here. Could be

```suggestion
            opts.append("--outdir=" + tool_output_dir)
```

Then change `temp_opts` to `opts` in `subproc_args`.

statick_tool/plugins/tool/cccc_tool_plugin.py
```diff
@@ -80,55 +90,85 @@ def scan(  # pylint: disable=too-many-branches
 
             logging.debug("%s", log_output)
 
-        if self.plugin_context and self.plugin_context.args.output_directory:
-            with open(self.get_name() + ".log", "wb") as flog:
-                flog.write(log_output)
+            if self.plugin_context and self.plugin_context.args.output_directory:
+                with open(self.get_name() + ".log", "wb") as flog:
```
denewiler, line 94: Since we are still running the tool once per file should we append to the log output? I did not test this.

```suggestion
                with open(self.get_name() + ".log", "ab") as flog:
```

tests/plugins/tool/cccc_tool_plugin/test_cccc_tool_plugin.py
```diff
@@ -291,3 +291,24 @@ def test_cccc_tool_plugin_scan_empty_calledprocesserror(mock_subprocess_check_ou
     )
     issues = ctp.scan(package, "level")
     assert not issues
+
+
+@mock.patch("statick_tool.plugins.tool.cccc_tool_plugin.xmltodict.parse")
+def test_cccc_tool_plugin_scan_filenotfound(mock_xmltodict_parse):
+    """
+    Test what happens when a CalledProcessError is hit (such as if cccc encounters an error).
```
denewiler, line 299:
```suggestion
    Test what happens when a FileNotFoundError is hit (such as if cccc has no output for a file).
```

denewiler (changes requested):
This worked great on all the examples I tested against. Nice work!

denewiler (approved):

PR #414, opened 2022-04-29
Fix support for user flags passed to the isort tool.

tdenewiler:


tdenewiler:
To test, you can use this new branch except copy in the version of `statick_tool/plugins/tool/isort_tool_plugin.py` from the main branch. The isort unit test will fail with that setup. If you use the version of `statick_tool/plugins/tool/isort_tool_plugin.py` from this new branch the unit test will succeed.

```shell
python -m pytest tests/plugins/tool/isort_tool_plugin/test_isort_tool_plugin.py
```

gregtkogut (approved):

Issue #422, opened 2022-06-07
Restructure tool plugins to not run if no files of their desired type are available.


xydesa:
There's no reason to run a tool plugin if there are no files available of the type that they operate on. Restructure the tool plugins so they do this.

PR #423, opened 2022-06-08
[#422] tool plugin refactor

xydesa:
#422
Restructure tool plugins to not run if no files of their desired type are available. In addition, I pulled out as much common code as made sense and converted as many plugins to the new structure as possible.

### DIFF
--- a/statick_tool/tool_plugin.py
+++ b/statick_tool/tool_plugin.py
@@ -41,7 +41,7 @@ class ToolPlugin(IPlugin):  # type: ignore
                 files += package[file_type]
 
         if files:
-            total_output = self.process_files(package, level, files, self.get_user_flags(level))
+            total_output = self.process_files(package, level, files, self.get_user_flags(level))  # pylint: disable=assignment-from-no-return
             if total_output is not None:
                 if self.plugin_context and self.plugin_context.args.output_directory:
                     with open(self.get_name() + ".log", "w", encoding="utf8") as fid:
--- a/statick_tool/tool_plugin.py
+++ b/statick_tool/tool_plugin.py
@@ -41,10 +41,8 @@ class ToolPlugin(IPlugin):  # type: ignore
                 files += package[file_type]
 
         if files:
-            total_output = (
-                self.process_files(  # pylint: disable=assignment-from-no-return
-                    package, level, files, self.get_user_flags(level)
-                )
+            total_output = (  # pylint: disable=assignment-from-no-return
+                self.process_files(package, level, files, self.get_user_flags(level))
             )
             if total_output is not None:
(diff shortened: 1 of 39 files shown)


statick_tool/tool_plugin.py
```diff
@@ -30,8 +30,35 @@ def get_tool_dependencies(cls) -> List[str]:
...
+                        for output in total_output:
+                            fid.write(output)
+
+                return self.parse_output(total_output, package)
+
+            return None
+
+        return []
+
+    def process_files(self, package: Package, level: str, files: List[str], user_flags: List[str]) -> Optional[List[str]]:
```
gregtkogut, line 57: Not sure what the best practice is for this, but pylint seems to prefer an explicit `None` return here, based on the failed build.    Are the black and pylint issues in the failed build real issues, or is that a case where we need to fix the build?

xydesa: Those are real issues. I wasn't sure what the best way to handle the desire for an explicit return statement here, given that they weren't needed in the past for overridden functions. but that may have been because the results weren't explicitly assigned to a variable in the past.

@denewiler any thoughts?

tdenewiler: I cloned your branch and I don't see how to fix that issue. The only other place I see it show up in Statick is in the same `tool_plugin` module, where we disable it. I recommend the same approach for this new one in `process_files`.

https://github.com/sscpac/statick/blob/main/statick_tool/tool_plugin.py#L82

xydesa: [255fba3](https://github.com/sscpac/statick/pull/423/commits/255fba353112268f619a63c6da380964e539aa93)

xydesa: Everything else resolved in: [2ac0e3b](https://github.com/sscpac/statick/pull/423/commits/2ac0e3b2d2bad9e6569d46e9ed52ae5fd15d88ec)

xydesa: > I cloned your branch and I don't see how to fix that issue. The only other place I see it show up in Statick is in the same `tool_plugin` module, where we disable it. I recommend the same approach for this new one in `process_files`.
> 
> https://github.com/sscpac/statick/blob/main/statick_tool/tool_plugin.py#L82

[2ac0e3b](https://github.com/sscpac/statick/pull/423/commits/2ac0e3b2d2bad9e6569d46e9ed52ae5fd15d88ec)

gregtkogut: Doh, I think black messed up your disable statement.

xydesa: ugh. that back and forth can be annoying. i think i really fixed it this time.

denewiler (approved):

Issue #432, opened 2022-07-05
Support auto-discovery


tdenewiler:
It would be nice to run all available discovery plugins and do auto-discovery of file types. Then, all applicable tools should run on discovered file types. Make it easier for all users, especially beginners.

denewiler:
Closed by #435.

PR #435, opened 2022-07-08
By default auto discover files of any type, and run any applicable tools on them. (#432)

xydesa:
Also only output to terminal by default. Adding do_nothing tool plugin.

#432


README.md
```diff
@@ -260,7 +260,11 @@ The `default` key lists the _level_ to run if no specific _level_ listed for a p
 
 The `packages` key lists _packages_ and override levels to run for those packages.
 
-With the built-in configuration files the default _profile_ uses `sei_cert` as the default _level_.
+With the built-in configuration files the default _profile_ uses `default` as the default _level_.
+This _level_ runs all available _discovery_ plugins, sets all available _tools_ to use their default flags,
+and only runs the _print_to_console_ reporting plugin (which outputs results to the terminal).
+
+With the built-in configuration files the another useful _profile_ uses the `sei_cert` _level_.
```
tdenewiler, line 267:
```suggestion
With the built-in configuration files another useful _profile_ uses the `sei_cert` _level_.
```

README.md
```diff
@@ -354,7 +358,7 @@ File Type | Extensions
 :-------- | :---------
 catkin    | `CMakeLists.txt` and `package.xml`
 C         | `.c`, `.cc`, `.cpp`, `.cxx`, `.h`, `.hxx`, `.hpp`
-CMake     | `CMakeLists.txt`
+CMake     | `CMakeLists.txt` and `.cmake`
```
tdenewiler, line 361:
```suggestion
CMake     | `CMakeLists.txt`, `.cmake`
```

statick_tool/rsc/profile.yaml
```diff
@@ -1,4 +1,4 @@
-default: sei_cert
+default: default
```
tdenewiler, line 1: It might be done already, but I would like it if we maintain an easy way to run the `sei_cert` level. Maybe by adding a `profile-sei-cert.yaml` file that has the default level as `sei_cert`? I want an easy upgrade path for users that were previously using the default level of `sei_cert`. This case should be supported by a unit test.

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+### Changed
+
+- Default behavior for statick will now run all available discovery plugins, and run all tool plugins where
```
tdenewiler, line 13:
```suggestion
- Default behavior for Statick will now run all available discovery plugins, and run all tool plugins where
```

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+  their desired source files are available, it will then output results only on the terminal. (#432, #435)
```
tdenewiler, line 14:
```suggestion
  their desired source files are available, then output results only on the terminal. (#432, #435)
```

tdenewiler:
I tested against a pure Python package and it works as described. I specified a user path and it used the default level specified in `profile.yaml` from that path. Then I did not specify a user path and it used the `default` level with default tool flags. This is a nice improvement.

statick_tool/rsc/profile.yaml
xydesa: That profile does exist already: https://github.com/sscpac/statick/blob/main/statick_tool/rsc/sei_cert.yaml
I'll add a unit test, should I also detail this in the changelog?

tdenewiler: I think it would be nice to add an upgrade path to the changelog. I can help if you want me to.

tests/plugins/tool/do_nothing_tool_plugin/test_do_nothing_tool_plugin.py
```diff
@@ -0,0 +1,66 @@
...
+from yapsy.PluginManager import PluginManager
+
+import statick_tool
+from statick_tool.package import Package
+from statick_tool.plugins.tool.do_nothing_tool_plugin import DoNothingToolPlugin
+from statick_tool.tool_plugin import ToolPlugin
+
+
+def setup_do_nothing_tool_plugin():
+    """Create and return an instance of the PyCodeStyle plugin."""
```
denewiler, line 13:
```suggestion
    """Create and return an instance of the do nothing plugin."""
```

statick_tool/rsc/profile.yaml
xydesa: [c4a97ab](https://github.com/sscpac/statick/pull/435/commits/c4a97ab1b2fff9d2f60582964598cfd96cb9e1ba)

denewiler (approved):

denewiler:
LGTM

PR #436, opened 2022-07-14
Adding "level" cli arg and unit tests. (#429)

### LINKED ISSUE #429: Support level flag
Right now a new profile file has to be created to specify the desired level to run Statick. We should add support for setting a `--level` flag that overrides the level specified in the profile file.

xydesa:
"level" cli arg overrides any levels specified by the profile used.

Closes #429


denewiler:
I tested this against a ROS1 workspace and a pure Python package.

LGTM

denewiler (approved):

PR #437, opened 2022-07-25
Removing unused comment.

gregtkogut:
This seems to fix an issue that is causing Actions to fail.  But I don't know what exactly changed that would change the behavior....but this change doesn't seem to fail self-check.

### DIFF
--- a/statick_tool/plugins/tool/cccc_tool_plugin.py
+++ b/statick_tool/plugins/tool/cccc_tool_plugin.py
@@ -193,7 +193,7 @@ class CCCCToolPlugin(ToolPlugin):
                     config[row[".ADA"]] = {
                         "warn": row["ada.95"],
                         "error": row[""],
-                        "name": row[None][3],  # type: ignore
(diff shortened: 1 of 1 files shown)


tdenewiler:
I was able to re-create this issue. I am using a Python virtual environment for testing. On this branch, run the following that results in an error.

```shell
$ pip install mypy==0.942
$ mypy --ignore-missing-imports --strict statick statick_tool/
statick_tool/plugins/tool/cccc_tool_plugin.py:196: error: Invalid index type "None" for "Dict[str, str]"; expected type "str"
Found 1 error in 1 file (checked 61 source files)
```

Upgrading to the latest ( currently `0.971`) fixes that issue on this branch.

```shell
$ pip install --upgrade mypy
$ mypy --ignore-missing-imports --strict statick statick_tool/
Success: no issues found in 61 source files
```

The changes you have in this branch are required for the latest PyPI version of mypy.

denewiler (approved):

PR #443, opened 2022-08-19
Add option to print timing information overall and for each plugin.

tdenewiler:


denewiler:
To test, run with

```shell
time ./statick . --output-directory /tmp/x --log info --profile self_check.yaml --timings
```

To speed things up you could limit the number of tools that are run.

```shell
time ./statick . --output-directory /tmp/x --log info --profile self_check.yaml --force-tool-list pycodestyle --timings
```

You should see a table printed after the reporting plugin is done. The overall time in the table should match very closely with the `real` time from the `time` utility, also printed out.

statick_tool/statick.py
```diff
@@ -530,7 +568,12 @@ def run_workspace(
                 mp_args.append((parsed_args, count, package, num_packages))
 
             with multiprocessing.Pool(parsed_args.max_procs) as pool:
-                total_issues = pool.starmap(self.scan_package, mp_args)
+                total_issues, all_timings = zip(  # type: ignore
```
tdenewiler, line 571: I had to ignore the type hint here, couldn't figure it out. Any insight or fixes would be nice.

statick_tool/statick.py
```diff
@@ -358,6 +379,9 @@ def run(
                 plugin.scan(package, level, self.exceptions)
```
xydesa, line 379: the first discovery package run is going to seem to take much longer than all the rest because of how the [`find_files`](https://github.com/sscpac/statick/blob/main/statick_tool/discovery_plugin.py#L40) function works (it runs exactly once per package assuming the plugin was updated to support this structure, which most discovery plugins have done). i think it'd be worth adding separate timing for this function if it's not too difficult.

denewiler: [d31d427](https://github.com/sscpac/statick/pull/443/commits/d31d427f23c2e6ff69466ae62c83624f75cf54c1)

Good ideas, I think it looks much better now.

xydesa (approved):

statick_tool/statick.py
```diff
@@ -332,6 +352,14 @@ def run(
         discovery_plugins = self.config.get_enabled_discovery_plugins(level)
         if not discovery_plugins:
             discovery_plugins = list(self.discovery_plugins)
+        # Get timing information for finding files for discovery plugins.
+        dummy_plugin = DiscoveryPlugin()
+        plugin_start = time.time()
+        dummy_plugin.find_files(package)
+        duration = format(time.time() - plugin_start, ".4f")
+        timing = Timing(package.name, "Find Files", "Discovery", duration)
```
xydesa, line 360: Nitpick: everything else in the timing table is lower case, this is the only thing with capitals in it.
```suggestion
        timing = Timing(package.name, "find files", "Discovery", duration)
```

xydesa: This would also be fine with me
```suggestion
        timing = Timing(package.name, "find_files", "Discovery", duration)
```

gregtkogut:
lgtm

gregtkogut (approved):

PR #445, opened 2022-08-25
Switch from apt to apt-get commands in CI Actions.

tdenewiler:
Better support for long-term backwards compatibility.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,34 +72,35 @@ jobs:
       - name: Install tools (Linux)
         if: matrix.os == 'ubuntu-18.04' || matrix.os == 'ubuntu-20.04' || matrix.os == 'ubuntu.22.04'
         run: |
-          sudo apt update
-          sudo apt install cccc
-          sudo apt install clang
-          sudo apt install clang-format
-          sudo apt install clang-tidy
-          sudo apt install cppcheck
-          sudo apt install libpcre3-dev
-          sudo apt install libperl-critic-perl
-          sudo apt install libxml2
-          sudo apt install libxml2-utils
-          sudo apt install shellcheck
-          sudo apt install uncrustify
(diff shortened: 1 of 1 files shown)


xydesa:
Where did you find this information? I'm curious and also it would be good to link here for people who want to know what prompted this change.

tdenewiler:
Good idea. I got the information from the apt tool itself. I added a comment to the action file. From the [apt manpage](https://manpages.ubuntu.com/manpages/kinetic/en/man8/apt.8.html):

>>>
SCRIPT USAGE AND DIFFERENCES FROM OTHER APT TOOLS

       The [apt](https://manpages.ubuntu.com/manpages/kinetic/en/man8/apt.8.html)(8) commandline is designed as an end-user tool and it may change behavior between
       versions. While it tries not to break backward compatibility this is not guaranteed either
       if a change seems beneficial for interactive use.

       All features of [apt](https://manpages.ubuntu.com/manpages/kinetic/en/man8/apt.8.html)(8) are available in dedicated APT tools like [apt-get](https://manpages.ubuntu.com/manpages/kinetic/en/man8/apt-get.8.html)(8) and [apt-](https://manpages.ubuntu.com/manpages/kinetic/en/man8/aptcache.8.html)
       [cache](https://manpages.ubuntu.com/manpages/kinetic/en/man8/aptcache.8.html)(8) as well.  [apt](https://manpages.ubuntu.com/manpages/kinetic/en/man8/apt.8.html)(8) just changes the default value of some options (see [apt.conf](https://manpages.ubuntu.com/manpages/kinetic/en/man5/apt.conf.5.html)(5)
       and specifically the Binary scope). So you should prefer using these commands (potentially
       with some additional options enabled) in your scripts as they keep backward compatibility
       as much as possible.
>>>

xydesa (approved):

PR #447, opened 2022-09-12
Add tabulate to install_requires in setup.py.

tdenewiler:


tdenewiler:
To test create a new virtual environment and do a local install of Statick. Then run Statick.

```shell
python3 -m venv tmp_venv
. tmp_venv/bin/activate
pip install -e .
statick --help
```

The main branch will currently fail because `tabulate` is not found. This branch will show the help output.

gregtkogut (approved):

PR #448, opened 2022-10-27
Update docker image

gregtkogut:
I only tested this with local docker build so far.   Wanted at least one level of sanity checking before building and tagging an image through the build-and-publish Action.


docker/Dockerfile
```diff
@@ -71,6 +67,10 @@ RUN python3 -m pip install --no-cache-dir --upgrade \
     pycodestyle \
     pydocstyle \
     pyflakes \
+    pylint \
+    pylint-django \
+    --ignore-installed pyyaml \
```
gregtkogut, line 72: This is because of a conflict with an OS-installed version.  It makes me wonder if we should just switch the image to using a venv for Python stuff.

tdenewiler: I like the idea of a Python venv in the Docker image. It's also nice to show in a Docker image how we install and execute things locally.

xydesa: i think using a venv in the docker image makes a lot of sense.

tdenewiler:
LGTM.

To get into the Docker container I ran

```shell
docker build -t foo . && docker run -it foo
```

For the pylint version I see

```shell
# pylint --version
pylint 2.15.5
astroid 2.12.12
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0]
root@968b8a221dbf:/#
````

**Review by tdenewiler — APPROVED** [review]

**Review by denewiler — APPROVED** [review]

**Review by denewiler — CHANGES_REQUESTED** [review]
Node is failing to install in test workflow. I will help fix it. Un-approving for now.

**Comment by gregtkogut** [comment]
Updated to use venv.  

To verify the new image, you can pull it and run locally.  

Also I create a private version of statick-action that uses the new image, and here is some output: 
https://github.com/gregtkogut/statick-action/actions/runs/3407241432/jobs/5666685464

It has an issue with statick-tooling as I write this.

File: docker/requirements.txt [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+pydocstyle
+pyflakes
+pylint
+pylint-django<2.0
+statick
+statick-md
+tabulate
+xmltodict
+yamllint
+yapsy
```
gregtkogut, line 22: It is less than ideal to have an exact copy of a file one directory up here.  But it's needed, at the moment, due to how Docker build contexts works.  I create an issue already to upgrade the use of the docker-build-and-push action to a new version that allows specifying multiple build contexts.   I tried to update it in this PR, but after spending 30 minutes fiddling with it, created the new issue to deal with it separately.

gregtkogut: I take this back.  It turns out that we want the requirements for the Docker image to be different than the one for statick proper, because we're adding the full suite of statick modules for use in the image.

gregtkogut:
I've resolved the issue the md-tooling, and statick-action is now completing successfully using this new image:  https://github.com/gregtkogut/statick-action/actions/runs/3424574285/jobs/5704399365

gregtkogut:
The rebased branch passes all checks.

denewiler (approved):

denewiler:
Changes look good and local testing showed the Docker image has desired tools, this time with Python tools available after activating venv in `/opt/venv`.

PR #454, opened 2022-11-29
Fix ci

gregtkogut:
This fixes the major CI issues that were blocking successful completion of CI.  

This PR can be used to now work on secondary issues and code coverage issues.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -14,34 +14,34 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
 
-      - uses: actions/cache@v1
+      - uses: actions/cache@v3
```
gregtkogut, line 28: This gets rid of a ton of deprecation warnings when viewing Action results, e[.g. as seen in this run](https://github.com/sscpac/statick/actions/runs/3563665572).   I went ahead and updated all versions that weren't throwing warnings just as general maintenance.

denewiler:
I tested locally, changes look good to me.

We have been having a discussion offline about the overuse of ignoring some type hint warnings from mypy but there is no consensus on whether there is a better approach yet. We can make a new issue if we find a better approach. Some options we are looking at are using the `@abstractmethod` decorator for empty body methods that are inherited by other classes, and having abstract classes return different types than derives versions of the functions.

denewiler (approved):

PR #460, opened 2022-12-20
Run pylint against all Python files at once. Support parallel processing flag of pylint.

tdenewiler:


xydesa (approved):

xydesa:
Nit: I like the idea of updating the changelog with each ticket. instead of all at once at the end. feel free to ignore this if you want.

statick_tool/rsc/config.yaml
```diff
@@ -136,12 +136,14 @@ levels:
         flags: ""
       pyflakes:
         flags: ""
+      # Disabling duplicate code as it is giving unreliable false positives when running pylint against multiple files
```
gregtkogut, line 139: Nit: It 
```suggestion
      # Disabling duplicate code as it is giving unreliable results when running pylint against multiple files
```

gregtkogut: Suggestion is because all the `duplicate-code` results I've seen it give were seemingly valid findings of duplicate code.  We'd consider them "false positives" because using the same 4 `import` lines in multiple files isn't something we'd consider bad.  But it is indeed finding duplicated lines between files - just not consistently.

statick_tool/statick.py
```diff
@@ -83,6 +83,18 @@ def set_logging_level(args: argparse.Namespace) -> None:
         logging.root.setLevel(log_level)
         logging.info("Log level set to %s", args.log_level.upper())
 
+    @classmethod
+    def set_cpu_count(cls, arg: str) -> int:
+        """Set correct number of CPU cores to use."""
```
gregtkogut, line 88:
```suggestion
        """Set number of CPU cores to use."""
```

statick_tool/statick.py
```diff
@@ -83,6 +83,18 @@ def set_logging_level(args: argparse.Namespace) -> None:
...
+        max_cpus = multiprocessing.cpu_count()
+        desired = int(arg)
```
gregtkogut, line 90: Nit, but `arg` is a vague argument name.  Can this be called something like `num_cpus`?

tdenewiler: [4514d04](https://github.com/sscpac/statick/pull/460/commits/4514d0465224ad043ccff760ffcdd769eb818388)

tests/plugins/tool/pylint_tool_plugin/test_pylint_tool_plugin.py
```diff
@@ -70,6 +72,21 @@ def test_pylint_tool_plugin_scan_valid():
     assert len(issues) == 2
 
 
+def test_pylint_tool_plugin_scan_valid_max_cpu_cores():
+    """Integration test: Make sure the pylint output hasn't changed when using max CPU cores."""
+    max_procs = multiprocessing.cpu_count()
```
gregtkogut, line 77: I usually see things called `max_procs` referring to a process count, not a cpu count.  Though since presumably pylint is spawning one process per cpu, I guess it's still valid.   I'm fine as-is, just pointing that out.

tdenewiler: Since the top-level argument for Statick is `--max-procs` I'm going to leave this as-is for now. If starting from scratch I would agree with you and change to `max_cpus` or similar.

gregtkogut (approved):

PR #461, opened 2022-12-21
Updated changelog for v0.9.2 release.

gregtkogut:


denewiler (approved):

denewiler (changes requested):
We also need to update the version number in statick_tool/__init__.py.

denewiler (approved):

PR #469, opened 2023-01-11
Move flake8 configuration max-line-length option to correct location.

tdenewiler:
When running pytest a warning was given about this. The max-line-length
was being set in the pytest section of our configuration (tox.ini). Moving it to
the flake8 section fixes the warning and makes more sense logically.


tox.ini
```diff
@@ -3,14 +3,14 @@ envlist = py37, py38, py39, py310, py311
...
 [pytest]
-flake8-max-line-length = 9000
 norecursedirs = .tox build docs examples
 
 # To work with black some items must be ignored.
 # https://github.com/psf/black#how-black-wraps-lines
 [flake8]
 exclude = .tox,tests/plugins/tool/black_tool_plugin/valid_package/format_errors.py,tests/plugins/tool/isort_tool_plugin/valid_package/*.py
 ignore = E126, E203, E231, E241, W503, W504
+max-line-length = 9000
```
xydesa, line 13: do you know if the length is set so large so that we only use `black` for checking line length? I couldn't tell based on the history of the file. It looks like it came from this PR: https://github.com/sscpac/statick/pull/54
I think it would be good to add a comment about why this value was chosen (if we know).

PR #471, opened 2023-01-14
Fix code climate reporting plugin bug.

tdenewiler:
Do not write output when an output directory is not specified. Fixes a
path where Statick would crash.


statick_tool/plugins/reporting/code_climate_reporting_plugin.py
```diff
@@ -125,6 +125,12 @@ def write_output(self, package: Package, level: str, line: str) -> bool:
         if not self.plugin_context:
             return False
 
+        if "output_directory" not in self.plugin_context.args:
```
xydesa, line 128: two things:
1. I feel like if output directory is not specified, but the configuration says to write an output file, then the output file should go in the current directory, not fail to be written. (line 36 of this file gets from the config whether or not to write an output file).
2. the json reporting plugin does not have these checks either.

tdenewiler: New commits cause code climate and json reporting plugins to write to current directory if the plugins are active for a level and no output directory is specified. That behavior is described in the README now as well. The unit tests were updated to test for this behavior.

Note that reporting plugins that exist outside of the main Statick repository should be checked that they follow the desired behavior.

xydesa: Tested the code-climate reporting plugin locally and the output file behavior works as desired.

xydesa (approved):

gregtkogut:
I used the command `statick . --level threshold --log info --timings`  (after adding code_climate as a reporting plugin, and it crashed in main, but not in this branch. 

LGTM.

gregtkogut (approved):

tdenewiler:
Thanks for coming up with testing instructions @gregtkogut!

PR #473, opened 2023-01-19
Add new tool plugin for the ruff tool.

tdenewiler:
Includes new `ruff` level to do only Python discovery and run `ruff`
tool. Ignores line-too-long warnings since those are better dealt with
by the `black` tool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ max            |                | 223    | 127    | 64
 
 - Process all source files at once with tools that support passing in a list of files, instead of invoking each tool
   per file. (#470)
+- New tool plugin for the [ruff](https://github.com/charliermarsh/ruff) tool.
 
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


tdenewiler:
I tested with:

```shell
python3 -m venv ruff-venv
. ruff-venv/bin/activate
pip install .
statick . --level self_check --log info --timings
```

The tool runs and no issues are reported. To see issues empty out the flags in the ruff level configuration. Then you should see some "line too long" issues. I ignored "line too long" ruff warnings because we use black for that and the two tools seem at odds with each other over what is proper and acceptable.

gregtkogut:
Code review looks good.  Just need to do test it out live...

gregtkogut:
> I tested with:
> 
> ```shell
> python3 -m venv ruff-venv
> . ruff-venv/bin/activate
> pip install .
> statick . --level self_check --log info --timings
> ```
> 
> The tool runs and no issues are reported. To see issues empty out the flags in the ruff level configuration. Then you should see some "line too long" issues. I ignored "line too long" ruff warnings because we use black for that and the two tools seem at odds with each other over what is proper and acceptable.

gregtkogut:
I followed the directions above, but don't see evidence that `ruff` is running

tdenewiler:
That's because I gave the wrong command to test, sorry about that. In the venv use the `ruff` level instead of the `self_check` level. To see lint too long issues use `ruff` level with the flags empty.

```shell
statick . --level ruff --log info --timings
```

As of now you should not see the `ruff` tool run as part of the `self_check` level.

gregtkogut (approved):

PR #474, opened 2023-01-29
Update version and CHANGELOG for new release: 0.9.3.

tdenewiler:

PR #475, opened 2023-01-29
Change json reporting plugin output file suffix from .json to .statick.json.

tdenewiler:


statick_tool/plugins/reporting/json_reporting_plugin.py
```diff
@@ -90,15 +90,10 @@ def write_output(self, package: Package, level: str, line: str) -> bool:
...
             logging.error("Unable to create output directory at %s!", output_dir)
             return False
-        output_file = os.path.join(output_dir, package.name + "-" + level + ".json")
-        next_output_file = os.path.join(
+
+        output_file = os.path.join(
             output_dir, package.name + "-" + level + ".statick.json"
         )
-        logging.warning(
-            "Output filename will change in statick v0.9.0 from %s to %s",
```
xydesa, line 98: should add in the changelog that this has taken place.

tdenewiler: You're right. I have a separate PR for a release (#474). I couldn't figure out how to have this go into the CHANGELOG until after that release. Or, we can choose to make this PR part of that release, in which case I can put this into the CHANGELOG, merge this PR, then merge that into #474. Any preferences?

xydesa: i think i'm ok either way, but since the other PRs are open, i think we should try to get them merged first.

xydesa:
Tested locally by adding this reporting plugin to the self_check profile and running that, the file created by this plugin has the new `.statick.json` extension.

xydesa (approved):

PR #477, opened 2023-01-30
In tex level configuration only use the tex discovery plugin.

tdenewiler:


statick_tool/rsc/config.yaml
```diff
@@ -228,6 +228,8 @@ levels:
         flags: "--ignore E501"
 
   tex:
+    discovery:
```
xydesa, line 231: should we update the change log to say this bug was fixed?

xydesa (approved):

xydesa (approved):

PR #478, opened 2023-01-31
Add file that causes UnicodeDecodeError when read into exceptions module.

tdenewiler:
Create new unit tests for exceptions module to test error and ignore packages configuration.


tdenewiler:
I found this issue when running the `rstcheck` tool plugin against the https://github.com/PointCloudLibrary/blog repository. I updated the `documentation` level to have the rst discovery plugin and the rstcheck tool plugin run.

An alternative way to test is to checkout the new test files into the main branch and you should see the unit test for exceptions fail. With this dev branch the unit tests pass.

If you want more details on how to test I can put them here.

xydesa (approved):

PR #479, opened 2023-02-01
Format test files.

tdenewiler:
- Ran `black` and `docformatter` against all the Python test files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Ran `black` and `docformatter` against all the Python test files.
+
+  ```shell
+  cd tests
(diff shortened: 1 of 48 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Ran `black` and `docformatter` against all the Python test files.
+
+  ```shell
+  cd tests
+  find . -name test_\*.py -exec black {} \;
```
xydesa, line 15: is there a way to remove just these files from the exception set for the `tests` directory?

tdenewiler: Not that I know of. My concern with this PR is that we are not going to check it in the future and it could turn into a periodic task. I felt that was better than nothing though.

xydesa (approved):

PR #481, opened 2023-02-12
Add support for multi-line yaml syntax.

tdenewiler:
* Use multi-line yaml for tool flags in default configuration file.
* Tool flags are now put on a single line with
    leading/trailing quotes and whitespace removed.
    This change is required for the flags passed to
    the make tool. Make requires more stringent
    formatting for flags. A single-line string is
    sufficient.
* Add unit tests to check that the config module can handle
    multi-line yaml syntax for tool flags.


tdenewiler:
I tested the configuration file changes against a ROS package, [node_example](https://github.com/tdenewiler/node_example). Assuming you have that in a ROS1 workspace, and you have sourced the `devel` space, then do this.

```shell
python3 -m venv venv
. venv/bin/activate
cd <statick/dev/branch>
pip install .
cd <node_example_ws_root>
statick src/node_example/ --output-directory /tmp/x --log info --timings --level <level>
```

I tried all the levels in the default `config.yaml` file that is installed with Statick: `sei_cert`, `threshold`, `objective_minus_pylint`, `objective`, `objective_with_format_minus_pylint`, `rosm_threshold`, `rosm_objective`, `self_check`, `ultimate`, `ruff`.

All of the commands completed successfully. I noticed that the `cmake` discovery plugin discovered make targets each time, and the make tool successfully ran, usually taking the longest of all tool plugins to complete.

statick_tool/config.py
```diff
@@ -157,7 +157,13 @@ def get_tool_config(
         self, plugin: str, level: str, key: str, default: Optional[str] = None
     ) -> Optional[str]:
         """Get tool flags to use for a plugin at a certain level."""
-        return self.get_plugin_config("tool", plugin, level, key, default)
+        tool_flags = self.get_plugin_config("tool", plugin, level, key, default)
+        # Make sure the flags are on a single line string and remove double-quotes and
+        # whitespace that might have been added from yaml multi-line syntax.
+        if tool_flags is not None:
+            tool_flags = " ".join(tool_flags.split()).strip('"').strip()
```
xydesa, line 164: why is removing `"` now required?

tests/config/rsc/multi-line-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+          "-W2
+          --ignore
+          DESCRIPTION_BOILERPLATE,
+          DESCRIPTION_MEANINGLESS,
+          GLOBAL_VAR_COLLISION,
+          LINK_DIRECTORY,
+          LITERAL_PROJECT_NAME,
+          TARGET_NAME_COLLISION"
+      cpplint:
+        flags: >
```
xydesa, line 17: Since we're claiming support for multi-line yaml, I think this file should use multiple different methods of multi-line breaks to make sure we test as many as we know about.
Alternatively, we should change the claim to just the method we know works and is tested.

statick_tool/config.py
tdenewiler: When we have all the flags on a single line and read them in there are no double quotes around the flags (when they are represented as a string in Statick code). The steps in `" ".join(tool_flags.split())` add the double quotes if the flags are specified using multi-line yaml syntax.

xydesa: I'm wondering why this part is needed `.strip('"')`

tdenewiler: That removes the leading and trailing `"` characters that are added when the flags are specified using multi-line yaml syntax. If the flags are specified on a single line then the `"` characters do not get added and the `.strip('"')` is a no-op.

tdenewiler: Note that the only tool that is sensitive to the leading/trailing `"` is `make`. That tool needs very specific formatting of flags that no other tools seem to care about.

xydesa (approved):

PR #483, opened 2023-04-08
Pin version of docformatter to be compatible with black. Update docstrings.

tdenewiler:


statick_tool/plugins/discovery/ros_discovery_plugin.py
```diff
@@ -29,6 +29,7 @@ def deep_get(
 
         Copied from:
 
+
```
xydesa, line 32: this second blank line is weird, but if it was needed to pass docformatter i'm fine with it.

tdenewiler: I agree. This is what comes of docformatter and black with our current dependency versions.

xydesa (approved):

PR #487, opened 2023-06-21
Drop support for Python 3.7 due to end-of-life on 27 June 2023.

tdenewiler:


xydesa:
Does it matter if we do this separately from the plugin repos?

xydesa (approved):

tdenewiler:
I'm not certain if removing a classifier for a specific distribution in setup.py has any effect on PyPI or pip install. My best guess is that the classifier is only used for documentation purposes and will not have any effect. Users on Python 3.7 systems will still be able to pip install statick and the statick plugins and use them. It is not until we make changes to the Statick code that are not supported by Python 3.7 that users will see any issues.

One piece of evidence to back that up is the wheel found on PyPI at https://pypi.org/project/statick/#files. The built distribution is statick-0.9.4-py3-none-any.whl. To me that implies Statick will be installed on all versions of Python 3 and on any architecture (working is a separate matter).

PR #494, opened 2024-02-03
Fix groovylint tool issues.

tdenewiler:
- Install lark as test dependency.
- Run tool against all source files at once instead of one at a time.
- Handle parsing issues when `CodeNarcServer` errors are present.


tdenewiler:
Opened #495 to have groovylint fixes and black formatting.

Issue #496, opened 2024-02-15
Encourage virtual environment for local development.


tdenewiler:
In the installation section of the README we should do more to encourage using a Python virtual environment before running `pip install statick`. We could also show how to do local installs from a cloned repository with `pip install -e .` in a venv. And discuss that a venv with multiple Statick repos needs to use `pip install .` (although I'm not sure of the root cause for why `pip install -e .` does not work with multiple Statick repos).

PR #498, opened 2024-02-16
[#496] Add Python virtual environment instructions to the README.

tdenewiler:


README.md
```diff
@@ -69,12 +69,25 @@ Statick is a plugin-based tool with an explicit goal to support external, option
 
 ## Installation
 
-Statick requires Python 3.6+ to run, but can be used to analyze Python2 projects, among many other languages.
+Statick requires Python 3 to run, but can be used to analyze Python 2 projects, among many other languages.
 
-The recommended install method is
+The recommended install method is to create a Python virtual environment and install from there.
```
xydesa, line 74:
```suggestion
The recommended install method is to create a Python virtual environment and install there.
```
I'm not sure if `from there` or `to there` is better, but I kinda like taking out that word and leaving it as `install there`.

xydesa (approved):

PR #499, opened 2024-02-24
Add short flag names

qdang922:
Added short flags "--output-drectory" as "-o" and "--user-paths" as "-u". Optest and black formatting ran successfully.

### DIFF
--- a/tests/plugins/tool/black_tool_plugin/valid_package/format_errors.py
+++ b/tests/plugins/tool/black_tool_plugin/valid_package/format_errors.py
@@ -1,1 +1,1 @@
-x =  "some string",
+x = ("some string",)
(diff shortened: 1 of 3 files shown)


tests/plugins/tool/black_tool_plugin/valid_package/format_errors.py
```diff
@@ -1 +1 @@
-x =  "some string",
+x = ("some string",)
```
denewiler, line 1: I think this change needs to be reverted. We don't want to run `black` on this file as we have intentional errors to trigger issues for unit tests. I expect a unit test will fail due to this change.

denewiler (approved):

PR #503, opened 2024-07-27
Fix command used to run ruff tool.

tdenewiler:
In v0.5.0 ruff requires use of `ruff check` instead of `ruff`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ pydocstyle
 pyflakes
 pylint
 pylint-django<2.0
-ruff
+ruff>=0.0.237
 statick
 statick-md
 tabulate
--- a/statick_tool/plugins/tool/ruff_tool_plugin.py
+++ b/statick_tool/plugins/tool/ruff_tool_plugin.py
@@ -25,26 +25,9 @@ class RuffToolPlugin(ToolPlugin):
         self, package: Package, level: str, files: List[str], user_flags: List[str]
     ) -> Optional[List[str]]:
         """Run tool and gather output."""
-        flags: List[str] = []
-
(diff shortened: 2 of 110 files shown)


statick_tool/plugins/tool/ruff_tool_plugin.py
```diff
@@ -26,23 +26,35 @@ def process_files(
...
+            )
+        except subprocess.CalledProcessError as ex:
+            logging.debug("Exception in first check of ruff, need to add check verb")
+            logging.debug(ex)
+            flags.append("check")
+        except OSError as ex:
+            logging.warning("Couldn't find ruff executable! (%s)", ex)
+            return None
+
+        flags += user_flags
```
xydesa, line 47: did `check` exist before v0.5.0? It seems like it did from at least [v0.1.0](https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#010)

I think we should just always use `ruff check` instead of this complicated pre-check

tdenewiler: Good find. I agree. I'll update.

xydesa: we could add a >0.0.237 to the setup.py ruff requirement: https://github.com/sscpac/statick/blob/main/setup.py#L59

requirements.txt
```diff
@@ -16,7 +16,7 @@ pydocstyle
 pyflakes
 pylint
 pylint-django<2.0
-ruff
+ruff>=0.0.237
```
xydesa, line 19: I think this change should be added to `setup.py` as well: https://github.com/sscpac/statick/blob/main/setup.py#L59

xydesa (approved):

PR #506, opened 2024-11-06
Adding statick-tex and json-merger to statick docker image

xydesa:


docker/requirements.txt
```diff
@@ -4,7 +4,9 @@ black
 cmakelint
 cpplint
 deprecated
-docformatter
+# Pin due to formatting conflict with black.
+# https://github.com/PyCQA/docformatter/issues/94
+docformatter==1.5.1
```
xydesa, line 9: I'm not sure if this is still required, but it's in the main `requirements.txt` file in the directory above, so I added it here too.

denewiler (approved):

PR #517, opened 2025-01-21
Add plugins from external repositories.

tdenewiler:


tdenewiler:
Before doing a new release we need to update the dependencies for the external statick plugin repositories. I think the next release should be a minor bump to 0.12.0, and we should pin the external plugin repositories to use `statick<0.12`.

tdenewiler:
I was able to get the history of the plugins, but I did not get the history of the unit tests. If that is important, I can go back and try again. The changes I made to unit tests were to change some integers to strings for issue asserts.

xydesa:
> I was able to get the history of the plugins, but I did not get the history of the unit tests. If that is important, I can go back and try again. The changes I made to unit tests were to change some integers to strings for issue asserts.

i'm ok with just the history for the plugins

tdenewiler:
Closing this PR. I will make a new branch with a better attempt to capture all git history.

PR #519, opened 2025-01-24
Add plugins from external repositories.

tdenewiler:


tdenewiler:
Before doing a new release we need to update the dependencies for the external statick plugin repositories. I think the next release should be a minor bump to 0.12.0, and we should pin the external plugin repositories to use statick<0.12. Pinning the dependencies that way will avoid potential conflicts between modules. An example is pip install statick statick-md, and then the user would have the markdown discovery module installed from two different packages.

tdenewiler:
I used these steps to get all the history.

https://blog.billyc.io/how-to-copy-one-or-more-files-from-one-git-repo-to-another-and-keep-the-git-history/

gregtkogut (approved):
LGTM

README.md
```diff
@@ -882,6 +909,49 @@ black statick statick_tool tests
 docformatter -i --wrap-summaries 88 --wrap-descriptions 88 <file>
 ```
 
+### Additional Installation
+
+#### Documentation Plugins
```
xydesa, line 914:
```suggestion
#### Plugins for NPM Tools
```
some of the tools mentioned here are not documentation related. I think this heading is more accurate.

xydesa (approved):

PR #522, opened 2025-02-07
Fix lizard plugin

tdenewiler:
Pin lizard to v1.17.14 as a workaround to the tool crashing with warning flag.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "importlib_metadata",
     "importlib_metadata",
     "isort",
-    "lizard",
+    "lizard==1.17.14",
```
xydesa, line 34: It looks like this fix was released in v1.17.18 already: https://github.com/terryyin/lizard/commit/2229380054a9657bc3f025d776e5e0f40e9a800a

should we change this to <=1.17.14 or >=1.17.18?

tdenewiler: I think unpinning the version is enough.

xydesa (approved):

PR #529, opened 2025-02-15
Change types of line number and severity in Issues from string to int.

tdenewiler:


src/statick_tool/plugins/reporting/json.py
```diff
@@ -51,10 +50,10 @@ def report(
             for issue in value:
                 issue_dict = OrderedDict()
                 issue_dict["fileName"] = issue.filename
-                issue_dict["lineNumber"] = issue.line_number
+                issue_dict["lineNumber"] = str(issue.line_number)
```
xydesa, line 53: is this really needed? json should support int types

tdenewiler: There could be a better way to address this. With `statick . -o /tmp/x --level self_check --log info` I get the following mypy issues.

```shell
Tool mypy: 2 unique issues
  /home/thomas/src/statick/statick/src/statick_tool/plugins/reporting/json.py:53: mypy:assignment: Incompatible types in assignment (expression has type "int", target has type "str")  [5]
  /home/thomas/src/statick/statick/src/statick_tool/plugins/reporting/json.py:56: mypy:assignment: Incompatible types in assignment (expression has type "int", target has type "str")  [5]
```

xydesa: based on my reading here, I think we could safely replace `OrderedDict` with a regular `dict` here and keep these as ints (I believe that issue is coming from using `OrderedDict`).

xydesa:
I'm getting a weird error when i try to test this branch in a uv venv:
```shell
(statick_ws) ~/code/github/statick_ws/tdenewiler-statick {list-dict-type $%} $ statick -h
  File "/home/axydes/code/github/statick_ws/.venv/bin/statick", line 4, in <module>
    from statick_tool.statick import main
ImportError: bad magic number in 'statick_tool': b'\x03\xf3\r\n'
```

src/statick_tool/plugins/reporting/json.py
xydesa: nevermind, that didn't change anything, but we can use a union type to have either str or int:
https://stackoverflow.com/questions/48013561/how-to-use-static-type-checking-using-dict-with-different-value-types-in-python

xydesa:
```python
                issue_dict: OrderedDict[str, Union[str, int]] = OrderedDict()
```

xydesa: [9db6d31](https://github.com/sscpac/statick/pull/529/commits/9db6d318d40903c5c1921862654b53c9f3dc98c3)

tdenewiler:
I'm thinking of replacing all `Union` type hints with the newer syntax, e.g., `str | None`, `str | int`. There are not that many instances: `grep -rn Union src/statick_tool/`.

xydesa (approved):

PR #530, opened 2025-02-15
Switch type hints from typing module Dict and List to built-in dict and list.

tdenewiler:


xydesa:
You missed an instance:
- https://github.com/sscpac/statick/blob/main/src/statick_tool/package.py#L3

Issue #490, opened 2024-01-22
Tool versions log option.


tdenewiler:
Add an option (`--show-versions`?) to print out the list of tool plugins and their versions. Have the tool plugin implement a default `get_version()` method that returns `Unknown`. Where possible in the individual tool plugins implement a `get_version()` method that gets the version using the Python API or a subprocess call. Return the version as a string. Then make a table like

Tool | Version
----- | -------------
black | 1.2.3
pylint | 4.5.6
cmake | 7.8.9

Having a default implementation will allow external plugin repos and tools that do not have an easily accessible version to come up with a solution later.

xydesa:
Added in #532

PR #532, opened 2025-03-11
Add command line interface arguments to show tool versions. (#490)

xydesa:
New arguments to show all tool versions, or just the ones enabled for the current run.

### DIFF
--- a/src/statick_tool/plugins/tool/black.py
+++ b/src/statick_tool/plugins/tool/black.py
@@ -21,7 +21,9 @@ class BlackToolPlugin(ToolPlugin):
         """Return a list of file types the plugin can scan."""
         return ["python_src"]
 
-    def get_binary(self) -> str:
+    def get_binary(  # pylint: disable=unused-argument
+        self, level: Optional[str] = None, package: Optional[Package] = None
+    ) -> str:
         """Get tool binary name."""
         return "black"
 
--- a/src/statick_tool/plugins/tool/black.py
+++ b/src/statick_tool/plugins/tool/black.py
@@ -25,7 +25,7 @@ class BlackToolPlugin(ToolPlugin):
         self, level: Optional[str] = None, package: Optional[Package] = None
     ) -> str:
         """Get tool binary name."""
-        return "black"
+        return self.get_name()
 
     def process_files(
         self, package: Package, level: str, files: list[str], user_flags: list[str]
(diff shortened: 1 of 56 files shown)


src/statick_tool/statick.py
```diff
@@ -48,6 +50,10 @@ def main() -> None:  # pragma: no cover
     if parsed_args.timings:
         print(tabulate(timings, headers="keys", tablefmt="pretty"))
 
+    if parsed_args.show_all_tool_versions or parsed_args.show_run_tool_versions:
+        tool_versions = statick.get_tool_versions()
+        print(tabulate(tool_versions, headers="keys", tablefmt="grid"))
```
xydesa, line 55: i like this table format `grid` better for tool versions. and I like the `pretty` table format better for timings. otherwise i'd use the same format.

src/statick_tool/plugins/tool/dockerfile_lint.py
```diff
@@ -15,18 +15,31 @@ class DockerfileULintToolPlugin(ToolPlugin):
 
     def get_name(self) -> str:
         """Get name of tool."""
-        return "dockerfile-lint"
+        return "dockerfile_lint"
```
xydesa, line 19: do you think this change will trip anyone up? it'll change what's in our output json. but i like the change.

src/statick_tool/plugins/tool/black.py
```diff
@@ -21,6 +21,12 @@ def get_file_types(self) -> list[str]:
         """Return a list of file types the plugin can scan."""
         return ["python_src"]
 
+    def get_binary(  # pylint: disable=unused-argument
```
tdenewiler, line 24: For all the `get_binary()` methods that do `return self.get_name()`, can you make this the method in `statick_tool.tool_plugin` module? Then you can remove it from all the tool plugins like this one.

xydesa: [66c84cc](https://github.com/sscpac/statick/pull/532/commits/66c84cc28d5261a96eafd4fabdfc0ddfdef2e4b2)

src/statick_tool/statick.py
tdenewiler: The `grid` table format looks good to me.

denewiler (approved):

PR #534, opened 2025-03-18
Adding a Devcontainer

xydesa:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,91 @@
+FROM ubuntu:24.04
```
xydesa, line 1: I was hoping to use the statick docker image as a basis, but the last published one was too old, and based on ubuntu 20.04.

So I  opted to create a new one.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+
+LABEL "name"="Statick DevContainer"
+LABEL "version"="0.0"
+LABEL "repository"="https://github.com/sscpac/statick"
+LABEL "homepage"="https://github.com/sscpac/statick"
```
kogut, line 6: These should be changed.  I'm about to open a PR that updates URL references project-wide.

xydesa: [b8a0d66](https://github.com/triboelectric/statick/pull/534/commits/b8a0d663e9c83ebc311058e693976c8f023bfa8d)

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+---
+services:
+  devcontainer:
+    build:
+      context: ../  # Start the context at the project root
+      dockerfile: .devcontainer/Dockerfile
+    volumes:
```
tdenewiler, line 7: Should we add a comment here that a user could mount their own code (from their host system) into a new volume? Then they could run Statick using Ubuntu 24.04-based tools on their code. If so, it would be good to test that use case. And, we should install more VS Code extensions for the languages we support with discovery. I'm fine if all of that is a follow-on issue where we can have more discussion and testing.

.devcontainer/scripts/dev.sh
```diff
@@ -0,0 +1,59 @@
...
+        cd $APP_HOME || exit
+        black src/statick_tool/ tests/
+        find src/statick_tool/ -name \*.py -exec docformatter -i --black {} \;
+        ;;
+    help)
+        show_help
+        ;;
+    qa)
+        cd $APP_HOME || exit
+        statick . -o /tmp/x -u statick_config/ --profile self_check.yaml --log info
```
tdenewiler, line 30: I prefer the approach of using `--level` instead of `--profile` these days. That results in one less required file anytime you want to add a custom configuration in your own package.

```suggestion
        statick . -o /tmp/x -u statick_config/ --level self_check --log info
```

tdenewiler (commented):
Note, when the dev container is active and I run `dev test` I'm seeing 2 tests fail. The tests are for `eslint` and `stylelint`. Those tools have been giving us problems for a long time on how to install and test them. I think it's worth taking a little time to try to fix them in the dev container. I don't know if the Dockerfile has to be updated, or the tests/plugins themselves.

```shell
=========================================================== short test summary info ===========================================================
FAILED ../tests/plugins/tool/eslint_tool_plugin/test_eslint_tool_plugin.py::test_eslint_tool_plugin_scan_valid_with_issues - TypeError: object of type 'NoneType' has no len()
FAILED ../tests/plugins/tool/stylelint/test_stylelint_tool_plugin.py::test_stylelint_tool_plugin_scan_valid_with_issues - TypeError: object of type 'NoneType' has no len()
```

tdenewiler:
This is a great addition!

.devcontainer/docker-compose.yml
tdenewiler: After talking more offline, we decided we want this dev container to be for developing Statick. The rest of this is outside the scope of what we want to maintain. It would be possible for other developers to use this dev container or our main Docker image to do their own development, but that will be up to them.

xydesa:
I don't know why, but i cannot tab complete `push origin` inside the devcontainer.

.devcontainer/scripts/dev.sh
```diff
@@ -0,0 +1,59 @@
...
+        cd $APP_HOME || exit
+        black src/statick_tool/
+        find src/statick_tool/ -name \*.py -exec docformatter -i --black {} \;
+        ;;
+    help)
+        show_help
+        ;;
+    qa)
+        cd $APP_HOME || exit
+        statick . -o /tmp/x -u statick_config/ --level self_check --log info
```
tdenewiler, line 30:
```suggestion
        statick . -o /tmp/x --level self_check --log info
```

When I run `dev qa` I see this error.

```shell
# dev qa
ERROR:root:Could not find user path statick_config/!
```

tdenewiler:
> I don't know why, but i cannot tab complete `push origin` inside the devcontainer.

I am able to tab complete remotes and branches with the `git push` command.

.devcontainer/scripts/dev.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Function to display help information.
+show_help() {
+  echo "Statick Developer Commands:"
+  echo "  dev format     - Format all Python code with black and docformatter."
+  echo "  dev help       - Show this help message."
+  echo "  dev qa         - Run linting and static analysis tools."
```
kogut, line 8: Be more explicit that this is running statick on itself?   Nit.

kogut:
Did limited testing, but it built find, and all the custom commands worked as they should.

kogut (approved):

PR #535, opened 2025-03-18
Updated organization references.

kogut:
I left one documentation reference I wasn't sure what to do since the repo doesn't exist in the new space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,6 @@ Statick is a plugin-based tool with an explicit goal to support external, option
     - [Discovery Plugins](#discovery-plugins)
     - [Tool Plugins](#tool-plugins)
     - [Reporting Plugins](#reporting-plugins)
-    - [External Plugins](#external-plugins)
   - [Customization](#customization)
     - [User Paths](#user-paths)
     - [Custom Profile](#custom-profile)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Statick
 
-![Unit Tests](https://github.com/sscpac/statick/workflows/Statick/badge.svg)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -532,20 +532,6 @@ Alternatively, Jenkins can be configured with quality gates to fail if a thresho
...
-Nearly all of these have been merged into the main repository, but they can serve as examples of how to make custom packages.
-
-Plugin Name      | Repository Location
-:--------------- | :-----------------------------------------------
-statick-fortify  | <https://github.com/soartech/statick-fortify>
-statick-md       | <https://github.com/sscpac/statick-md>
-statick-planning | <https://github.com/tdenewiler/statick-planning>
-statick-tex      | <https://github.com/tdenewiler/statick-tex>
-statick-tooling  | <https://github.com/sscpac/statick-tooling>
-statick-web      | <https://github.com/sscpac/statick-web>
```
kogut, line 547: Wasn't sure what to do here.  I could re-instate the ones in the `tdenewiler` space?

tdenewiler: I think we should leave those. There is a note at the top of the section that they are stale and have been merged into the main repository. Leaving them will provide a reference for how to generate a set of third party, possibly proprietary plugins, which is still a use case we want to support. The only thing missing with them is I couldn't get code coverage working with the stdlib-based plugin approach. Other than that, the external plugins do work.

kogut: I added back in the tenewiler ones.   You're saying we should leave the ones still in the `sscpac` org. too?

tdenewiler (approved):

README.md
tdenewiler: I'm fine as-is. You could add the `sscpac` ones back in. I think the `statick-fortify` one should be removed since it doesn't follow the stdlib plugin approach so it's not a good example.

xydesa (approved):

axydes (approved):

PR #537, opened 2025-03-23
Add git-completion script to dev container.

tdenewiler:


tdenewiler:
With this script as part of the dev container, I can tab complete git commands. Without this script, I cannot tab complete git commands.

axydes (approved):
Changes look good and work for me.

Issue #525, opened 2025-02-12
Move `statick-action` into this main repo


xydesa:
https://github.com/sscpac/statick-action

Hopefully use a subdirectory for most of the files, and only leave the minimum required at the top level.

kogut:
@xydesa and @tdenewiler , when testing out the statick-action self-test I picked up the below findings.


I'm happy to fix them, but just wanted your quick input on fix vs. suppress, etc.


```
Tool shellcheck: 22 unique issues
  /github/workspace/.devcontainer/config/.git-prompt.sh:1: shellcheck:SC2148: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive. [5]
  /github/workspace/.devcontainer/config/.git-prompt.sh:102: shellcheck:SC2155: Declare and assign separately to avoid masking return values. [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:113: shellcheck:SC2321: Array indices are already arithmetic contexts. Prefer removing the $(( and )). [1]
  /github/workspace/.devcontainer/config/.git-prompt.sh:137: shellcheck:SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting). [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:140: shellcheck:SC2178: Variable was used as an array but is now assigned a string. [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:141: shellcheck:SC2178: Variable was used as an array but is now assigned a string. [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:144: shellcheck:SC2178: Variable was used as an array but is now assigned a string. [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:144: shellcheck:SC[229](https://github.com/kogut/statick/actions/runs/14020335350/job/39251436013#step:3:231)5: Expansions inside ${..} need to be quoted separately, otherwise they match as patterns. [1]
  /github/workspace/.devcontainer/config/.git-prompt.sh:147: shellcheck:SC2128: Expanding an array without an index only gives the first element. [3]
  /github/workspace/.devcontainer/config/.git-prompt.sh:239: shellcheck:SC2086: Double quote to prevent globbing and word splitting. [1]
  /github/workspace/.devcontainer/config/.git-prompt.sh:324: shellcheck:SC2162: read without -r will mangle backslashes. [1]
[...]

Issue #507, opened 2024-11-15
If files are ignored via an exception, remove them during the discovery/find_files process


xydesa:
https://github.com/sscpac/statick/blob/main/statick_tool/discovery_plugin.py#L41

This will remove them super early on and hopefully prevent processing lots of unwanted files.

kogut:
In addition to ignoring, either his issue or a follow-on could also modify the workflows so that workflow artifacts don't end up in the repo tree unless necessary - for example mypy caches or tox venvs.

axydes:
I tried moving the `filter_file_exceptions_early` call into the `discovery_plugin.find_files` function (which is only called once), but the timings of all discovery plugins (and that specific function) all increased after this change.

we already call that function in the `scan` function of most discovery plugins. That might be as good as we can get.

tdenewiler:
One thing I tried in a forked branch (that I threw away) was to switch our `os.walk()` call in `find_files()` with `os.scandir()`. You can call `os.scandir()` recursively, since it returns the files and directories in the current directory it is looking at. See https://stackoverflow.com/questions/33135038/how-do-i-use-os-scandir-to-return-direntry-objects-recursively-on-a-directory for an example. The benefit I was hoping for was to pass a list of exceptions into `find_files()`. then along with each `.is_dir(entry)` call I added a check to make sure the entry was not in our list of exceptions.

The result I got was the `scandir` approach was 2-4 times slower than our current implementation. That's why I deleted my branch.

PR #541, opened 2025-04-17
[#525] Move Statick action into main repo.

kogut:
The test is verifying the self-test action works.  

I had to manually update-and-push to the Docker image used because the `0.12.0` version was really running statick `0.10.0`. 

I also had to disable pyright for the action because `0.12.0` obviously doesn't support that plugin.   Though I like having to chain user paths because it's good self-documentation for prospective users of the action on how to refer to their own custom configs.  If there's agreement, I'll make an issue to update the pipeline so that pull requests - not just releases - trigger image builds-and-pushes.  With the PR versions tagged as `main` and the release ones with the semver.  That way things like tool additions won't break the self-test action.  This same issue would likely update the Dockerfile so that is uses `pyproject.toml` instead of the `requirements.txt` file in the docker subdirectory.....and in fact installs statick directly from a clone instead of from pip.    We might want a separate Dockerfile that pulls from pip just as a sanity check that the pip version isn't broken due to some dependency change.


.github/workflows/self-test.yml
```diff
@@ -0,0 +1,22 @@
...
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          path: "statick"
```
kogut, line 16: This tripped me up for longer than it should have.  I didn't realize that the checkout action doesn't result in a named repository.  It just places the contents of the repo directly under `/github/workflow`.   Statick appeared to run fine, except that it would think the package name was "workspace" instead of "statick," and a couple of our self-check exceptions specify package name "statick."    Took me a while to figure that out - maybe we should have statick report the package name it's using at the INFO level?    So this `path` parameter tells checkout to place the contents of the repo under a directory called `statick.`.

src/statick_tool/plugins/discovery/cmake.py
```diff
@@ -170,13 +170,14 @@ def process_output(  # pylint: disable=too-many-locals
             output: The output to parse.
             package: The package to update.
         """
-        # pylint: disable=anomalous-backslash-in-string
+        # pylint: disable=anomalous-backslash-in-string, line-too-long
```
kogut, line 173: Not sure why the other self-check isn't finding these two lines.....

axydes:
> The test is verifying the self-test action works.
> 
> I had to manually update-and-push to the Docker image used because the `0.12.0` version was really running statick `0.10.0`.
> 
> I also had to disable pyright for the action because `0.12.0` obviously doesn't support that plugin. Though I like having to chain user paths because it's good self-documentation for prospective users of the action on how to refer to their own custom configs. If there's agreement, I'll make an issue to update the pipeline so that pull requests - not just releases - trigger image builds-and-pushes. With the PR versions tagged as `main` and the release ones with the semver. That way things like tool additions won't break the self-test action. This same issue would likely update the Dockerfile so that is uses `pyproject.toml` instead of the `requirements.txt` file in the docker subdirectory.....and in fact installs statick directly from a clone instead of from pip. We might want a separate Dockerfile that pulls from pip just as a sanity check that the pip version isn't broken due to some dependency change.

1. I like the idea of PRs creating docker images for the self-test workflow, but it would be better if they were tagged with the PR number, not `main`
2. I think the official release versions of the action would be good to pip install from pypi in the docker image, and the PR versions can do a git clone. I have not idea how hard that would be to differentiate, but just wanted to point out what I think would be best.

.github/workflows/self-test.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: Statick
```
axydes, line 2: I think we should rename this to `Test Action`
```suggestion
name: Test Action
```
I can't tell it apart from the other `test.yml` workflow in the github website.

And maybe rename the `test.yml` workflow to `Self Test` on line 2.

kogut:
> 1. I like the idea of PRs creating docker images for the self-test workflow, but it would be better if they were tagged with the PR number, not `main`
> 2. I think the official release versions of the action would be good to pip install from pypi in the docker image, and the PR versions can do a git clone. I have not idea how hard that would be to differentiate, but just wanted to point out what I think would be best.

Good point on both.  Yes, there'd be a PR version tagged with the PR #, but was thinking also a post-merge `main` that could be used in a nightly test (or similar) on the current `main` branch of the repo?    So there'd always be a functional action for the `main` branch.

axydes:
Do you know if this will result in a "view action in marketplace" header in the main repo once everything is released?

kogut:
> Do you know if this will result in a "view action in marketplace" header in the main repo once everything is released?

I need to refresh myself on the details of how actions are published to the marketplace.   I'll look at that tomorrow.    But, yes, I think once we do a new "action release," that'd be updated.   That's a separate process from our own package release process.

Roughly speaking, after this is merged, I think we'd:

1) Do a draft release to the Marketplace.
2) Switch our workflow to use the Marketplace version instead of referring to our own repo.
3) After some testing, move the draft to fully released.

axydes:
The action version of the self check workflow isn't finding as many files as the regular workflow:
- regular: https://github.com/triboelectric/statick/actions/runs/14525311758/job/40755457791?pr=541#step:17:20
- action: https://github.com/triboelectric/statick/actions/runs/14525311755/job/40755457752?pr=541#step:3:128

axydes:
> The action version of the self check workflow isn't finding as many files as the regular workflow:
> 
> * regular: https://github.com/triboelectric/statick/actions/runs/14525311758/job/40755457791?pr=541#step:17:20
> * action: https://github.com/triboelectric/statick/actions/runs/14525311755/job/40755457752?pr=541#step:3:128

nevermind, it's ending up with the same number of files after filtering. the difference in file count before filtering probably has to do with the venv in the regular workflow.

kogut:
> > The action version of the self check workflow isn't finding as many files as the regular workflow:
> > 
> > * regular: https://github.com/triboelectric/statick/actions/runs/14525311758/job/40755457791?pr=541#step:17:20
> > * action: https://github.com/triboelectric/statick/actions/runs/14525311755/job/40755457752?pr=541#step:3:128
> 
> nevermind, it's mostly ending up with the same number of files after filtering. the difference in file count before filtering probably has to do with the venv in the regular workflow.
> 
> the regular workflow does have more dockerfiles getting scanned than the action one for some reason.

It might be because the action is using `0.12.0` and the other uses `main` - maybe there are some differences?   Just for fun I'll try a `main` version of the docker image in my own repo.

kogut:
For the Dockerfile stuff, we have two issues not related to this PR.

1)  It is reporting, IMO, false positives:

```
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/tool/dockerfilelint.py
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/tool/dockerfile_lint.py
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/tool/__pycache__/dockerfilelint.cpython-312.pyc
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/tool/__pycache__/dockerfile_lint.cpython-312.pyc
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/discovery/dockerfile.py
INFO:root:Dockerfile found: /home/greg/src/statick/src/statick_tool/plugins/discovery/__pycache__/dockerfile.cpython-312.pyc

```

2) If you specify that statick's output be placed in the package itself, it find even more false positives:

```
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/tool/dockerfilelint.data.json
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/tool/dockerfile_lint.data.json
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/tool/dockerfile_lint.meta.json
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/tool/dockerfilelint.meta.json
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/discovery/dockerfile.meta.json
INFO:root:Dockerfile found: /home/greg/src/statick/statick-output/statick-self_check/.mypy_cache/3.12/statick_tool/plugins/discovery/dockerfile.data.json
```

This suggests a larger issue that statick's own output is being considered for discovery at all, which I don't think it should be.  We could maybe self-add an early-file exception for whatever the output is?

For the issue with calling anything with "dockerfile" in a Dockerfile, we could add a stricter test based on this from the docker docs: _"A Dockerfile must begin with a FROM instruction"_

I can create a separate issue for this stuff if there's agreement.

kogut:
I think there's something wrong with discover in the the tox version.  It's reporting over 8000 Python files.  That's simply wrong.  

My guess is it could be doing something like running discovery in sphinx output from a prior step.  Or some other output, like mypy's.   If we could verify that, we should probably purify the directory before running `statick` on it.

tdenewiler:
> I think there's something wrong with discovery in the the tox version. It's reporting over 8000 Python files. That's simply wrong.
> 
> My guess is it could be doing something like running discovery on sphinx output from a prior step. Or some other output, like mypy's. If we could verify that, we should probably purify the directory before running `statick` on it.

When are you seeing this happen? Sometimes when I run statick after I run tox, I see a lot of Python files. I assume (and I think I verified years ago) that many of the files are in the tox virtual environments. We have #507 that would/should fix that, but I haven't been able to get a working solution for that so far (I should put my notes on that issue...). I'm wondering if that is where you are seeing 8000 Python files?

tdenewiler:
> For the Dockerfile stuff, we have two issues not related to this PR.
>
>...
> 
> And this isn't output from a prior run, but output from the current run already being picked up during discovery.
> 
> This suggests a larger issue that statick's own output is being considered for discovery at all, which I don't think it should be. We could maybe self-add an early directory exception for whatever the output path is?
> 
> For the issue with calling any filename with "dockerfile" in it a Dockerfile, we could add a stricter test based on this from the docker docs: _"A Dockerfile must begin with a FROM instruction"_ Alternatively, we could exclude any filename with `dockerfile` in it that also has an extension. I've never seen a true Dockerfile with a filename extension.
> 
> I can create a separate issue for this stuff if there's agreement.

I agree that both of those are issues. I like your suggested approach to Dockerfile discovery, where we could look to see if `A Dockerfile must begin with a FROM instruction` is present before adding the file.

For running Statick against it's own output, #507 could help. But, if we can't solve that soon, we can at least remove those files in between steps in CI. 

One thing most plugins try to do is only write output if an output directory is specified. An example from `shellcheck`:

```python
        if self.plugin_context and self.plugin_context.args.output_directory:
            with open(self.get_name() + ".log", "w", encoding="utf8") as fid:
                fid.write(output)
```

I would want to test more, but we might want to prepend `self.plugin_context.args.output_directory` to the filename in the `open` call. That might be a separate bug, and not necessarily solve the issue you discovered.

One more idea is to add `**/.mypy_cache/**/*.json` to our standard file exceptions.

action/README.md
```diff
@@ -0,0 +1,66 @@
+# statick-action
+
+![Statick](https://github.com/triboelectric/statick-action/workflows/Statick/badge.svg)
```
tdenewiler, line 3: We'll need to update this URL.

.github/workflows/rsc/config.yaml
tdenewiler, line 1: Is it possible for us to symlink this file to the one in the main statick src/ path?

kogut: They're slightly different for now.   The workflow version has pyright removed because for now the docker image uses statick `0.12.0` which has no pyright plugin.   On my other comments I addressed possible ways to keep the action in sync with repo changes like new plugins by pushing new docker images on PR and PR merges.

kogut:
> > I think there's something wrong with discovery in the the tox version. It's reporting over 8000 Python files. That's simply wrong.
> > My guess is it could be doing something like running discovery on sphinx output from a prior step. Or some other output, like mypy's. If we could verify that, we should probably purify the directory before running `statick` on it.
> 
> When are you seeing this happen? Sometimes when I run statick after I run tox, I see a lot of Python files. I assume (and I think I verified years ago) that many of the files are in the tox virtual environments. We have #507 that would/should fix that, but I haven't been able to get a working solution for that so far (I should put my notes on that issue...). I'm wondering if that is where you are seeing 8000 Python files?

Yes, I phrased it wrong.    We are leaving tox artifacts in the workspace tree before doing self-check.   I did a quick check of one sample of the tox matrix, and it leaves 15914 files to be considered for discovery when there are 1821 in the repo (525 branch).  (https://github.com/kogut/statick/actions/runs/14550199690/job/40820457118#step:17:14) I'll take a look at #507 above after this one.   One way is to "ignore" them per #507, but it's also quite easy to move where tox puts its venv to keep the repo pure in the first place.  But it might take a little work to make sure the workflow can find the artifacts correctly.

.github/workflows/rsc/config.yaml
tdenewiler: I read that and didn't put 2 and 2 together. Those are good approaches.

action/README.md
kogut: Thanks for catching this.  I fixed the PR branch version, [viewable here](https://github.com/kogut/statick/blob/525-move-statick-action-into-this-main-repo/action/README.md). 

- [ ] Change to expected `main` version just before merge.

tdenewiler (approved):

axydes (approved):

Issue #510, opened 2025-01-03
Cppcheck SEI Cert flags.


tdenewiler:
The cppcheck tool has added support for checking more items listed in the SEI Cert guide on software safety. We need to update the plugin mapping file for cppcheck so Statick supports reporting of these issues.

- https://wiki.sei.cmu.edu/confluence/display/c/Cppcheck
- https://github.com/danmar/cppcheck/blob/a7707a457dd70ee539becea57622276618926926/addons/cert.py#L416

PR #545, opened 2025-04-30
[#510] Update mapping of cppcheck warnings to SEI CERT findings.

tdenewiler:


axydes:
Not necessary for this PR, but it would be nice to have documentation on how to update these. where you get the information from, etc.

axydes (approved):
Seems fine to me and does run.

tdenewiler:
> Not necessary for this PR, but it would be nice to have documentation on how to update these. where you get the information from, etc.

506fc89. Looking at updating those instructions, I think we could

1. rename `plugin_mapping` to `sei_cert_mapping` to make it easier to add different mappings in the future, and to be more clear about the current mapping
2. use the list of analyzers in the links to see if any of the tools are good candidates for Statick plugins

